PR #91, opened 2014-02-21
Allow fields to be added to Zones

dlh01:
This was inspired by a project I saw where editors could choose a page layout (1 big post and 3 small posts, or 4 medium posts, etc.) while managing posts in a zone.

### DIFF
--- a/css/fieldmanager.css
+++ b/css/fieldmanager.css
@@ -272,3 +272,10 @@ a.fm-delete:hover {
 .edit-tags-php .fm-media-button {
 	width: 200px;
 }
+
+/**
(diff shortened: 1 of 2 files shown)


xmattus:
In re the relationship between this and #140, this allows fields on which `add_term_form` is called to end up on zoninator's zone pages (zones = terms, of course). #140 adds fields that would normally be associated with posts (`add_meta_box`) to the draggable items within the zone. So they both change the zone manager page, but in unrelated ways.

mboynes:
This along with #140 should be moved to an add-on plugin, "Fieldmanager_Zoninator". Closing here.

PR #119, opened 2014-03-14
Prevent missing fields from throwing a nonce error

mboynes:
We will need to do this (or similar) across all contexts. I can add the other contexts to this PR, but wanted to run the gist of the solution by here first.

The problem I ran into is that if a fieldmanager field is added between the time that a user loads the new/edit post screen and the time that the user saves the draft or publishes the post, Fieldmanager will throw an exception about an invalid nonce. Fieldmanager currently mandates that the field is present in the POST data.

What this PR does is ensure that both the field data and nonce are present before proceeding. If the nonce is there but the data isn't, the field will be saved as empty. If the field is there but the nonce isn't, the normal nonce check will continue and fail.


mboynes:
This is no longer an issue.

PR #140, opened 2014-05-09
Zoninator context

xmattus:


xmattus:
Ok, tests added. I'm not sure how kosher the way I'm loading up the zoninator plugin is in the setUp method of the test class, so I'd appreciate any feedback on that. Otherwise I think this is finally ready to go.

mboynes:
This along with #91 should be moved to an add-on plugin, "Fieldmanager_Zoninator". Closing here.

xmattus:
Code now lives at https://github.com/alleyinteractive/fm-zoninator

PR #157, opened 2014-06-23
User forms

netaustin:
User-facing forms to include in themes, etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,47 +2,47 @@ sudo: false
 
 language: php
 
-php:
-    - 5.3
-    - 5.4
-    - 5.5
-    - 5.6
-
-env:
-    - WP_VERSION=master WP_MULTISITE=0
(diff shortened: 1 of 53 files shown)


mboynes:
This should not delete `Fieldmanager_Context_Page`, as others may be using that. We will deprecate it elsewhere first.

montchr:
I'm finally checking this out since `Fieldmanager_Context_Page` has been deprecated.

I've replaced all calls to `add_page_form()` in my project with `add_form()` and everything seems to work functionality-wise.

However, styles for the rendered form are a bit messed up:

image 2018-05-03 at 14 32 00

Here's what it should look like normally:

image 2018-05-03 at 14 33 35

Looks like that's happening because the parent of `.fm-group-label-wrapper` is `.fm-fm-group` instead of `.fm-group`. Not sure exactly why that's happening.

If someone resolves the merge conflicts I could take another look.

PR #174, opened 2014-07-27
Gallery field

danielbachhuber:
I think we might've discussed this before, but I don't see an existing thread for it.

How would you feel about a gallery field that used the Media Library to pick a selection of images? Using a repeating / sortable media field is a bit janky UX.


js/media/fieldmanager-media.js
```diff
@@ -20,44 +64,158 @@ $( document ).on( 'click', '.fm-media-button', function( event ) {
...
+
+		}
+
+	});
+
+	var query_args = {
+		'type': 'image',
+		'post__in': selectedImages,
+		'orderby': 'post__in',
+		'perPage': -1
```
danielbachhuber: Not sure how VIP is going to feel about this. Might need to add some transparent pagination.

danielbachhuber:
@mattheu Couple of things still:

* [ ] If I'm using an image field with `limit => 0`, as Fusion is, I'd expect I could use `collection => true` to turn all of those images into a gallery. The behavior as it exists in this PR is to implement repeatable collections, which introduces a new data structure.
* [ ] When I edit the collection, I'd expect it to open the edit gallery experience, not the UI to select images again.

@mboynes @netaustin still interested in this PR, or should we internalize first? what's the backlog like getting something onto VIP?

mattheu:
Currently you can switch from `limit = 0` and `collection=false` to `limit=false` and `collection=true` as the data structure is the same here. 

Also If you click a collection to edit - it does already open the edit gallery experience...

danielbachhuber:
Hmm. Maybe I didn't pull latest. I'll test again today

danielbachhuber:
My bad on testing — the issues are non-issues.

Because we need this now, we've moved it to a separate repo: https://github.com/fusioneng/fieldmanager-gallery

Happy to pick things up again when there's some direction from the maintainers on where we'd like this to go.

mboynes:
@danielbachhuber I'd love to finally merge https://github.com/fusioneng/fieldmanager-gallery into FM. Your thoughts? I'm inclined to keep it a separate field rather than add onto Fieldmanager_Media as you did initially. One potential issue is naming -- we don't want to break any sites where your plugin is installed.

danielbachhuber:
> Your thoughts?

I'm open to it.

> I'm inclined to keep it a separate field rather than add onto Fieldmanager_Media as you did initially.

Seems reasonable.

>  One potential issue is naming -- we don't want to break any sites where your plugin is installed.

I suspect this is a half dozen or less. We haven't really given it much promotion, and we only have one site. My bad not changing the name in the first place.

enrico-sorcinelli:
I just sent a new [PR](https://github.com/alleyinteractive/wordpress-fieldmanager/pull/506) with an updated Fieldmanager gallery version (with unit tests) in order to finally close this thread ;-)

mboynes:
Closing in favor of #506

PR #183, opened 2014-08-05
Save an empty term set for a field

willgladstone:
When a field has only_save_to_taxonomy as true and append_taxonomy as false, it does not write to the terms table with an empty set.  This is problematic when using the checkboxes field to set terms for a post.


bcampeau:
@mboynes @noah tests added if you want to re-review

PR #204, opened 2014-09-01
Add support for attributes in groups

mboynes:


mboynes:
Indeed, this breaks collapsibility. Fix forthcoming.

mboynes:
As initially written, this PR breaks some JS functionality (e.g. collapsing groups) because it adds a dom layer and the jquery targeting expects a specific hierarchy. That's easy enough to fix, but:

1. Assuming we move forward as-is, we should consider this a breaking change, since a theme or plugin might use similar targeting, following Fieldmanager's lead.
2. The question should be asked: is adding a wrapping div the appropriate implementation for group attributes? An alternative without adding DOM nodes would be to add them to `div.fm-group-inner`, though there are two downsides to that approach: (a) that only renders if the group has a label, so we'd have to always render that div and then conditionally add the `.fm-group-inner` class based on if there's a label or not, and (b) the attributes wouldn't wrap the tabs if there were any.

mboynes:
I made a commit that fixes the JS bugs, so this is prepared to move forward as-is if we determine that's the best bet. My two cents is that it is; I think adding the attributes to `.fm-group-inner` would be less useful and would offer minimal gains. Another option would have been to add the attributes to the `.fm-item` wrapper, but that would be less consistent with other fields' `attributes` implementation. This feels at least _somewhat_ consistent.

dlh01:
Re. backcompat, could the wrapping div be omitted if there are no attributes on the group, would you consider that a sufficient protection against calling this a breaking change, and would avoiding the breaking change be worth the maintenance of maintaining two DOM structures for groups?

mboynes:
That's a clever idea. It would certainly mitigate any backwards compatibility issues. So the question, as you highlight, becomes "would avoiding the breaking change be worth maintaining two DOM structures for groups?"

🤷

mboynes:
My hunch is that it's not worth it. Groups are already a tangled mess, let's not add another knot.

dlh01:
Moving some PRs with potentially breaking changes out of the 1.4 milestone so that it can be as easy as possible for sites to upgrade to 1.4, with those potential breaking changes being made available in a focused 1.5.

PR #206, opened 2014-09-01
Proposal for adding text before a field or group

mboynes:
This is an unfinished proposal for this. I like the idea, but I'm not sure on where exactly to add the text.


danielbachhuber:
Closing in favor of #443

PR #239, opened 2014-10-22
Media field: trigger change event after updating/removing hidden field value

mattheu:
It would be useful if a change event were triggered when a file/media field changes the value of the hidden inuput.

This replicates the event that would fire were a user to change the field value.

This is a pretty edge case requirement - but i'm using the fieldmanager fields to live update a preview, and currently it is not easy to detect whether this value has changed. By triggering this event manually, I can listen for this and update my preview in exactly the same way that I do with standard input fields.

I understand that my use case isn't really standard, however I think this is a fairly small change that makes sense to add.


mboynes:
Happy to merge this after it gets updated. Closing for now to clean house.

Issue #244, opened 2014-10-30
Select / Chosen type_ahead incompatible with Term datasource


danielbachhuber:
In our Post's term fields, we want to persist the assigned order so the first term assigned to a Post can be considered the "primary" term. Fieldmanager's Select Field with `type_ahead` and the term data source only works part of the way for this.

Using Chosen, I can assign terms to a post in arbitrary order. The terms are saved to the database correctly. However, because the `<select>` is built with `<option>` taxonomy terms in alphabetical order, my specified order is ignored for display (and lost when I save the post).

But, I think the bug is that `use_ajax` doesn't work with Chosen. For my term data source, I don't want to load all of the terms at once because it's a performance issue.

ankitrox:
This is still an issue where the order isgetting changed in chosen field. 
Are there any plans to fix this issue?

PR #251, opened 2014-11-06
Swap out Chosen for Select2

danielbachhuber:
* [x] Replace Chosen with Select2.
* [x] Update js/fieldmanager-select.js to do whatever is needed for updating Select2
* [ ] If AJAX data source is provided, use that.

See #244


kevin-brown:
> Ugh. Select2 doesn't support ajax datasources for <select>.

Might as well mention that as of 4.0.0, this is supported (and is the recommended default)

danielbachhuber:
We'll consider Select2 for Fieldmanager 2.0.0, but not going to change things for 1.0.0

PR #252, opened 2014-11-07
Restore AJAX data source for Chosen

danielbachhuber:
In doing so, we can avoid the performance issues with large taxonomies

See #244


danielbachhuber:
:football: not worth pursuing this. Select2 is the way to go.

PR #259, opened 2014-11-23
Term Meta cleanup

mboynes:
* phpdoc
* update methods to always return post meta functions
* coding standards


nschoenholtz:
The return value change seems so obvious in hindsight! And nice cleanup, especially awesome docs :100:

Issue #260, opened 2014-12-03
fm_get_context() can miscalculate for submenus of custom menus


dlh01:
I've got a patch for this issue and am working on a unit test.

When using, e.g., `fm_register_submenu_page( 'my_fields', 'my_custom_menu_slug', 'Meta Boxes' );`, the corresponding `fm_submenu_my_fields` hook doesn't fire.

dlh01:
First addressed in #263.

PR #263, opened 2014-12-05
Submenu contexts for custom menus; submenu context test

dlh01:
There are a few parts to this PR:

* 2f0d1d8, which makes sure that all FM submenus hooked at priority 10 are added
* b7c2645, which is the attempted fix for #260 
* bab112d and 982adfe, which introduce a function and tests for calculating the FM context
* 87434f1, which stops `fm_trigger_context_action()` from firing an action of `fm_`


mboynes:
Looks great, seems to work well. :+1:

Issue #241, opened 2014-10-27
Checkboxes should also support "display-if"


jitendraharpalani58:
I ran into a use-case of having a checkbox and a text field, where, text field should only be visible if checkbox is "checked". Because "display-if" doesn't work with checkboxes, i had to use drop-down list. It would be nice to make this enhancement, though.

PR #264, opened 2014-12-05
Display if support for Checkboxes

jitendraharpalani58:
#241

@danielbachhuber Works for me locally. Please verify..


goldenapples:
Just a friendly bump - this functionality would be nice to have, as I ran into this problem again today.

Issue #54, opened 2013-10-17
Style enhancements for Repeatable fields not in a Fieldmanager_Group


mboynes:
If you set `'limit' => 0` on a Fieldmanager_* object which isn't Fieldmanager_Group, the box title bar is absent. This is (I believe) the intended behavior, though it leaves some styling issues:

1. <del>The "drag icon" that appears isn't properly lined up with the label</del> *Edit: no longer an issue*
2. <del>The "(x)" icon isn't properly lined up with the label</del> *Edit: no longer an issue*
3. The field has `overflow:auto` set, which leaves a hint of a scroll bar
4. <del>It's worth noting that this is also subject to #43, where a label is necessary for the move and remove icons to appear.</del> *Edit: resolved in #203*

```
$fm = new Fieldmanager_Textfield( array(
	'name'           => 'repeatable_text',
	'limit'          => 0,
	'add_more_label' => 'Add another field',
	'sortable'       => true,
	'label'          => 'Text Field'
) );
$fm->add_meta_box( 'Repeatable Text Fields', 'demo-text' );
```

![screen shot 2013-10-17 at 11 37 46 am](https://f.cloud.github.com/assets/465154/1353217/1d8559a0-3742-11e3-8931-c33a7a6fe37f.png)

Issue #50, opened 2013-10-10
Some field types need max-width set in post sidebar


mboynes:
Specifically, text fields (and the like, e.g. date, password, autocomplete, etc.), textareas, richtextareas, and (potentially) select fields. This is in both MP6 and the legacy admin style.

![screen shot 2013-10-09 at 11 52 29 pm](https://f.cloud.github.com/assets/465154/1303604/a60f46d4-315f-11e3-885e-97389f0df153.png)

![screen shot 2013-10-09 at 11 52 45 pm](https://f.cloud.github.com/assets/465154/1303607/a9f3c568-315f-11e3-931f-09b31a57b3ef.png)

PR #273, opened 2014-12-21
Style updates

### LINKED ISSUE #48: Term Context has some styling bugs, especially with mp6
1. Radios, checkboxes are on their own line
screen shot 2013-10-09 at 11 43 36 pm
2. Radio buttons are huge on MP6
screen shot 2013-10-09 at 11 44 06 pm
3. Fields are 100% width, which isn't terrible but is inconsistent with the post context
4. Groups are too subtle on MP6
screen shot 2013-10-09 at 11 44 20 pm
5. Buttons are inconsistent with other contexts and are full width (they don't look like buttons, in fact)
screen shot 2013-10-09 at 11 44 38 pm
6. The "New Term" form is too narrow for the full Rich Text Editor, depending on the screen width. Any laptop will cut it off. Not sure if there's a solution to this, unfortunately.
screen shot 2013-10-17 at 10 55 00 am

mboynes:
Closes #48, #50, #54


nschoenholtz:
seems like a nice improvement, :+1:

PR #274, opened 2014-12-21
Force first_empty for repeatable selects

### LINKED ISSUE #150: Select field without 'first_empty' in a repeatable group adds iteration with each save
Using the following code:

```php
$fm = new Fieldmanager_Group( array(
	'name'           => 'repeatable_select',
	'limit'          => 0,
	'add_more_label' => 'Add another field',
	'label'          => 'Field',
	'children'       => array(
		'select_field' => new Fieldmanager_Select( 'Repeatable Field', array( 'options' => array( 'One', 'Two', 'Three' ) ) )
	)
) );
$fm->add_meta_box( 'Repeatable Select Fields', 'post' );
```

Every time you save a post, it will add another iteration of `repeatable_select` because `select_field` will never be empty.

Not sure what the solution is here. Perhaps force `first_empty` on repeatable select fields?

mboynes:
Also adds basic tests for select fields. Requires #255, so this is a placeholder until that gets merged.

Closes #150.

### DIFF
--- a/php/class-fieldmanager-select.php
+++ b/php/class-fieldmanager-select.php
@@ -94,7 +94,7 @@ class Fieldmanager_Select extends Fieldmanager_Options {
 		}
 
 		$opts = '';
-		if ( $this->first_empty ) {
+		if ( $this->is_repeatable() || $this->first_empty ) {
 			$opts .= '<option value="">&nbsp;</option>';
 		}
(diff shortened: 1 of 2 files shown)


nschoenholtz:
holy regular expressions batman! looks good to me, might see me ripping this off for rendering tests in the near future :+1:

PR #275, opened 2014-12-21
Update chosen to 1.3.0

mboynes:
This is a merge into #252


nschoenholtz:
Haven't had a chance to test it yet, but the code looks good :+1:

danielbachhuber:
:football:

mboynes:
We should probably update chosen regardless. I'll see if this can easily be forklifted to a new PR into `master`

PR #278, opened 2014-12-23
Added store_property to Fieldmanager_Datasource_User

bcampeau:
I ran into a use case (replacing the default Co-Authors Plus user mapping meta box) where it made more sense to store user_login than ID. Therefore, I've added 'store_property' which allows you to save 'user_login', 'user_email', 'user_nicename' or 'ID' (the default).

Reciprocals don't work with anything other than ID so it throws an error if you try to enable that and the code validates it in appropriate places as well. This is going into production on the site that required it, so I'd like to get this into master.


mboynes:
I noticed today that this datasource's phpdoc is all copypasta from datasource_post. When you finish this PR, can you fix all the phpdoc too?

bcampeau:
Not my copy pasta but sure.

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -58,8 +74,23 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
...
+    	switch ( $this->store_property ) {
+    		case 'ID':
+    			$field = 'ID';
+    			$value = intval( $value );
+    			break;
+    		case 'user_nicename':
+    			$field = 'slug';
+    			break;
+    		case 'user_email':
+    			$field = 'email';
```
bcampeau: See below.

bcampeau:
Comments addressed. Working on unit tests.

tests/php/test-fieldmanager-datasource-user.php
```diff
@@ -0,0 +1,163 @@
+<?php
+
+/**
+ * Tests the Fieldmanager Datasource User
+ *
+ * @group datasource
+ * @group post
```
mboynes: This should be `@group user`

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -44,22 +57,56 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
...
+			throw new FM_Developer_Exception( __( 'You cannot use reciprocal relationships with FM_Datasource_User if store_property is not set to ID.', 'fieldmanager' ) );
+		}
+	}
 
     /**
-     * Get a post title by post ID
+     * Get a user by the specified field.
      * @param int $value post_id
-     * @return string post title
+     * @return WP_User
```
mboynes: Looks like this should be returning a `string`, right?

mboynes: Or an `int`, I guess. Maybe worth just callign it `mixed` Regardless, worth adding a description that it returns whatever `display_property` is

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -156,4 +219,25 @@ public function get_edit_link( $value ) {
...
+    		case 'user_email':
+    			$value = sanitize_email( $value );
+    			break;
+    		default:
+    			$value = sanitize_text_field( $value );
+    			break;
+    	}
+    	
+    	return $value;
+    }
```
mboynes: Niiiiice. I would make this `protected` instead of private so classes extending this one can use it.

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -34,6 +34,19 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
...
+     * Store property. Defaults to ID, but can also be 'user_login', 'user_email',
+     * or 'user_nicename'.
+     */
+    public $store_property = 'ID';
+    
+    /**
+     * @var array
+     * Allowed store properties for validation.
+     */
+    private $allowed_store_properties = array( 'ID', 'user_login', 'user_email', 'user_nicename' );
```
mboynes: I think this should be `protected`

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -44,22 +57,56 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
...
+    		case 'user_login':
+    			$field = 'login';
+    			break;
+    	}
+    	
+    	// Sanitize the value
+    	$value = $this->sanitize_value( $value );
+        
+        $user = get_user_by( $field, $value );
         return $user ? $user->{$this->display_property} : '';
```
mboynes: Looks like some of the above lines might be space-indented -- they don't look quite right

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -98,44 +148,57 @@ public function get_ajax_action() {
...
-        if ( $field->data_type != 'post' || !$this->reciprocal ) return $values;
-        foreach ( $current_values as $user_id ) {
-            delete_user_meta( $user_id, $this->reciprocal, $field->data_id );
+        if ( $field->data_type != 'post' || ! $this->reciprocal || 'ID' != $this->store_property ) {
+        	return $values;
         }
+        
+        if ( ! empty( $current_values ) ) {
+	        foreach ( $current_values as $user_id ) {
+    	        delete_user_meta( $user_id, $this->reciprocal, $field->data_id );
```
mboynes: Indentation. Looks like some of this might be space-indented too.

php/datasource/class-fieldmanager-datasource-user.php
bcampeau: Updated

php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -44,22 +57,56 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
...
+		if ( ! in_array( $this->store_property, $this->allowed_store_properties ) ) {
+			throw new FM_Developer_Exception( sprintf( 
+				__( 'Store property %s is invalid. Must be one of %s.', 'fieldmanager' ),
+				$this->store_property,
+				implode( ', ', $this->allowed_store_properties )
+			) );
+		}
+		
+		if ( ! empty( $this->reciprocal ) && 'ID' != $this->store_property ) {
+			throw new FM_Developer_Exception( __( 'You cannot use reciprocal relationships with FM_Datasource_User if store_property is not set to ID.', 'fieldmanager' ) );
```
mboynes: Good catch

php/datasource/class-fieldmanager-datasource-user.php
bcampeau: Did

mboynes:
This is awesome! One request: looks like there's no test coverage for `$display_property`. Can you add a test which covers this? Might be worth adding a matrix of tests to cover all the possible `$store_property` and `$display_property` values. After that, :+1: :+1:

bcampeau:
I've added validation for display_property as well as tests now.

PR #282, opened 2014-12-30
fix so apply_filters() doesn't just result in 'Array'

joshkadis:


xmattus:
The only site using it is JTA, and I wouldn't really object to making it userland only. It was built fairly specifically to their needs and we might be retiring it soon anyway.

PR #284, opened 2014-12-31
Fix null case on new fields when only_save_to_post_parent => true

### LINKED ISSUE #283: PHP Notice or extra fields when Fieldmanager_Datasource_Post::$only_save_to_post_parent = true
This code will trigger a PHP notice:
```php
	'name'           => 'debug_datasource',
	'limit'          => 1,
	'datasource'     => new Fieldmanager_Datasource_Post( array(
		'only_save_to_post_parent' => true,
		'query_args' => array( 'post_type' => 'post' )
	) )
) );
$fm->add_meta_box( 'Post Datasource', 'post' );
```

And this code will result in two fields to start instead of 1:
```php
	'name'           => 'debug_datasource',
	'limit'          => 0,
	'datasource'     => new Fieldmanager_Datasource_Post( array(
		'only_save_to_post_parent' => true,
		'query_args' => array( 'post_type' => 'post' )
	) )
) );
$fm->add_meta_box( 'Post Datasource', 'post' );
```

Issue appears to be [here](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/datasource/class-fieldmanager-datasource-post.php#L262), as this always returns an array with an element. I think it needs to return an empty array if there are no values.

mboynes:
Resolves #283


mboynes:
I'm going to revisit the tests covering `only_save_to_post_parent` before merging this. It should have caught this issue, so I want to make sure it covers the rendering as well as the saving.

mboynes:
Possible dupe of #307, need to figure out which solution is better.

PR #285, opened 2015-01-02
Add option for vertical tab orientation

mboynes:


php/class-fieldmanager-group.php
```diff
@@ -38,10 +38,11 @@ class Fieldmanager_Group extends Fieldmanager_Field {
 	public $collapsed = FALSE;
 
 	/**
-	 * @var boolean
-	 * If true, render children in tabs.
+	 * Use tabbed groups. Currently supports "horizontal" or "vertical".
```
nschoenholtz: This comment's slightly misleading (false, horizontal, or vertical) and doesn't explain the deprecated true == horizontal. I'm actually cool with leaving out the latter I think, but the three possible states of the param should probably be explicit.

nschoenholtz:
It's basically all CSS! That's a compliment! Very cool. One comment about docs, :pizza: depending on how you feel about that.

PR #286, opened 2015-01-04
Add Fieldmanager_Context_Storable, to differentiate contexts which store data from contexts which don't

mboynes:
Some contexts may not load or store data, and it wouldn't need to implement the abstract methods in the base class `Fieldmanager_Context`. This PR adds a new class `Fieldmanager_Context_Storable`, which is an extension to `Fieldmanager_Context` and is meant to serve contexts which do load and store data.


netaustin:
:tulip: although i think this warrants manual testing of each context before merge

mboynes:
All contexts manually tested

PR #288, opened 2015-01-08
Add a label to the term meta post type

dlh01:
Although this post type is private, the label still appears on the Export page, because `$can_export` is not false, and it might appear elsewhere.

Without its own label, the post type is labeled "Pages."


mboynes:
Nice catch!

PR #298, opened 2015-01-20
fixed misnamed class for Fieldmanager_Context_Page

nschoenholtz:
I think this is going to be superseded by https://github.com/alleyinteractive/wordpress-fieldmanager/pull/157, but in the meantime I just want to fix this little bug so that I can update FM to the latest on an old project that uses `Fieldmanager_Context_Page`.


mboynes:
:+1:

Issue #18, opened 2013-06-03
Confusing data type for $values in Fieldmanager_Field::element_markup()


mboynes:
The PHPDoc says that `$values` should be an array. If a field is singular or empty, (I think) this is actually a string. If it's an empty repeating field or group, it's an empty string, but `count()` is run on it, resulting in 1 when it should be 0.

The entire chain of `$values` should be traced, and the data type(s) should be enforced so that it always contains the exact type of data we're expecting. This can be an array for repeating fields and a string otherwise, or it can always an array. I'd opt for the latter for the sake of predictability.

mboynes:
It is what it is.

PR #307, opened 2015-02-24
Fix data type issues with only_save_to_post_parent post datasources

mboynes:
Relates to #18


php/datasource/class-fieldmanager-datasource-post.php
```diff
@@ -228,7 +228,7 @@ public function presave( Fieldmanager_Field $field, $value, $current_value ) {
                 );
             }
             if ( $this->only_save_to_post_parent ) {
-                return array();
+                return '';
```
netaustin: make this `null` to match `Fieldmanager_Datasource_Term`

netaustin:
One little tweak and :horse

mboynes:
Possible dupe of #284, need to figure out which solution is better.

PR #310, opened 2015-03-02
Fix an issue with checkboxes

mboynes:


nschoenholtz:
:ballot_box_with_check:

PR #315, opened 2015-03-07
Fix bug with submenu slashes

### LINKED ISSUE #314: RichTextArea in a submenu context seems to get double escaped
In a submenu context, HTML content gets double escaped independently of the sanitization function.
```
<a href="\&quot;http://google.com\&quot;">Testing</a>, <strong>testing \'</strong>
```

I could swear I remember us encountering and fixing this issue before, something about WP Core options escaping that we don't can't alter, and having to unescape and let core escape the data... maybe?

Tim needs this fix by Wed 2015/03/18

mboynes:
Closes #314; addresses bug introduced in bdb9b82

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -507,6 +507,11 @@ function fm_sanitize_textarea( $value ) {
 }
 
 /**
+ * Stripslashes_deep for submenu data.
+ */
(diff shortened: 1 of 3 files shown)


nschoenholtz:
Nice, thanks! :pizza: :rocket:

PR #318, opened 2015-03-17
Exclude term meta post type from search

### LINKED ISSUE #317: Exclude term meta post type from search?
The fm-term-meta post type is currently not excluded from search by default. Is there a reason for this? Off the top of my head I can't think of a reason that we'd want the term meta to show up in search results by default..

jjeaton:
Fixes #317.


mboynes:
Great catch, and thanks for the PR! I'm going to add on to this, because several of these arguments are unnecessary.

mboynes:
Since I can't commit to this PR directly, I've opened #319. Thanks again!

PR #319, opened 2015-03-17
Remove unnecessary post type args

### LINKED ISSUE #317: Exclude term meta post type from search?

mboynes:
Replaces #318, closes #317. Props @jjeaton


jjeaton:
:+1:

bcampeau:
:horse: pending Travis

PR #323, opened 2015-04-02
Adding custom args to autcomplete ajax requests

bcampeau:
This adds a new parameter called `custom_args` that lets you create a handler that can append arbitrary data to an AJAX request. The assumption is that if you wanted to use it in a query, you would also implement your own `query_callback` for the datasources that support that. It's otherwise harmless and ignored.


mboynes:
:horse_racing:

PR #324, opened 2015-04-07
Display If: Match the checked value on load

matthewspencer:
For radio buttons `trigger.val();` returns the first element’s value, not the checked element’s value. This PR adds a check to see if there is a checked value when the trigger initializes.

#264 has a related check for checkboxes.


mboynes:
Great catch, thanks for fixing it!

matthewspencer:
@mboynes Thanks for merging this! :smile: Looking forward to having this available on VIP, how long does that usually take?

bcampeau:
VIP is currently in the process of reviewing a Fieldmanager update. This may make it into that review. I expect it to be deployed this week.

matthewspencer:
@bcampeau Rad. Thank you.

bcampeau:
@matthewspencer this is now live on VIP

PR #328, opened 2015-04-20
Only check user capability when attempting to save user meta

danielbachhuber:
Checking it all the time means users can't be a data source when the
user doesn't have `list_users`


danielbachhuber:
Closing in favor of #329

PR #329, opened 2015-04-20
Only check user capability when attempting to save user meta, attempt 2

danielbachhuber:


danielbachhuber:
Core lets editors assign authors without `list_users`, which means editors should be able to list and save users without necessarily having `list_users`.

danielbachhuber:
Genesis is 63f78832d7064a6dea8d6214222f830e0f4f2c36, which doesn't help much.

danielbachhuber:
@netaustin would love your thoughts on this when you have a moment.

danielbachhuber:
@mboynes @netaustin any thoughts on this?

mboynes:
This update makes good sense to me. I'm going to update the branch, and we might have to make one other update to allow `add_user_meta()` to be overwritten for environments that don't use user meta.

mboynes:
After reviewing the code in-depth, it looks like this was intentional -- that one shouldn't be allowed to save a field using the user datasource if that user isn't allowed to `list_users`. The original intent had nothing to do with the reciprocal value.

This is tricky. As it stands, it doesn't make much sense -- Fieldmanager allows you to see the list of users, it just doesn't allow you to save any data about them (to whatever the current context is). IMO, if a user cannot `list_users`, the field shouldn't render at all, because that ultimately gives them a list of users.

@danielbachhuber it seems like you have a situation where you want someone without the `list_users` capability to be able to use a field with the user datasource. Can you elaborate on your use-case, and what's your $0.02 about what I said above about not rendering the field at all?

danielbachhuber:
> it seems like you have a situation where you want someone without the list_users capability to be able to use a field with the user datasource. Can you elaborate on your use-case, and what's your $0.02 about what I said above about not rendering the field at all?

Correct. Our use case is that we have show pages like [Soccer Gods](http://fusion.net/show/soccer-gods/) with contributor modules like this:

image

A show page is a Show custom post type with some Fieldmanager fields. We'd like Producers (who can edit the show page) to be able to populate the contributor module without having the `list_users` capability. However, I've granted them `list_users` as an intermediate workaround and it's not a blocker.

Hiding the element wouldn't make sense in our case. Because our data is being saved to the Post context, I shouldn't think the `list_users` capability should have an impact on what data can be saved.

mboynes:
> Because our data is being saved to the Post context, I shouldn't think the list_users capability should have an impact on what data can be saved.

I agree that `list_users` should have no impact on what data can be saved to a post. Here's my point about the field rendering, in case that wasn't clear:

Let's say I'm a user who cannot `list_users`, but can create a post. On the new post screen, say there's a dropdown to select other contributors for the story. That dropdown is a list of users, which I can see, even though I lack the `list_users` capability. If it's intentional that I cannot see other users on the site, this is a loophole where I can.

There's another issue related to this code, which is that if reciprocal is set, the capability to write to the user field should probably be `edit_user`.

danielbachhuber:
> Let's say I'm a user who cannot list_users, but can create a post. On the new post screen, say there's a dropdown to select other contributors for the story. That dropdown is a list of users, which I can see, even though I lack the list_users capability. If it's intentional that I cannot see other users on the site, this is a loophole where I can.

Right, but core already permits this: https://github.com/alleyinteractive/wordpress-fieldmanager/pull/329#issuecomment-94578267

PR #336, opened 2015-06-05
Remove term meta rewrite rules

dlh01:
After #319, the term meta post type falls back to the default `rewrite => true`, which generates unused (I think) rewrite rules.


mboynes:
:horse:

dlh01:
Merging on the theory that https://core.trac.wordpress.org/ticket/32444#comment:87 is the likely culprit in that build's demise.

PR #357, opened 2015-06-29
Fix for serializable wp_filters

firejdl:
`Fieldmanager_Media::__construct()` was defining an action as a closure, which makes it so that `wp_filters` isn't serializable. This pull request moves the action into a class function instead.


nschoenholtz:
Apologies, looks like your travis tests were spoiled because we hadn't yet merged https://github.com/alleyinteractive/wordpress-fieldmanager/pull/355

nschoenholtz:
Pending travis (which I expect will pass now that I've restarted it after merging the other PR), I think this is good to go.

firejdl:
Thanks for tackling the Travis error, @nschoenholtz! :+1: looks like it's good to go now.

nschoenholtz:
Thanks @firejdl!

Issue #341, opened 2015-06-09
Quickediting adds zeros to quickedit box


mboynes:
See the upper-right corner in this screenshot:
![screenshot 2015-06-09 16 22 19](https://cloud.githubusercontent.com/assets/465154/8068729/bcca68d4-0ec3-11e5-912c-bb84dd023ee5.png)

Note that each time you open a quickedit box, a zero is added. Clearly a JS issue.

Here's the code used to generate that screenshot:

```php
add_action( 'fm_quickedit_post', function() {
	$fm = new Fieldmanager_TextField( array(
		'name' => 'demo',
	) );
	$fm->add_quickedit_box(
		'Basic Field',
		'post',
		function( $values ) {
			return ! empty( $values['demo'] ) ? $values['demo'] : '';
		}
	);
} );
```

PR #358, opened 2015-06-30
Remove errant "0" from ajax response in quickedit context

goldenapples:
This fixes the issue described in #341. 

Exiting after outputting the desired ajax response prevents WordPress from spitting out its ugly zeroes.


mboynes:
Thanks for tracking that down! Can you do me one quick favor before we merge this and convert the spaces before the `exit` to tabs?

goldenapples:
Ah, thanks, I didn't notice that. I've gotten so used to depending on `.editorconfig` to enforce tabs or spaces I forget to check that I'm set correctly locally.  Fixed now...

PR #359, opened 2015-07-01
Update draggable-bin so field name can have '-bin'

kshaner:
The input hidden element for draggable posts is linked by the id of the wrapper.

```
$field = new Fieldmanager_DraggablePost(array(
    'name' => 'dragged-posts',
    'repositories' => array(
        /* arguments */
    ),
    'bins' => array(
        'post-bins' => 'Bins'
    ),
));
```

The javascript fails because "post-bins-bin".replace('-bin', '') returns "posts-bin" and not the id of the input 'post-bins'.  This patch makes sure it's only replaces -bin at the end of the id.

### DIFF
--- a/js/fieldmanager-draggablepost.js
+++ b/js/fieldmanager-draggablepost.js
@@ -33,7 +33,7 @@
 	function populateHiddenElements() {
 		$('.post-bin').each(function(i) {
 			var post_ids = [];
-			var input_name = $(this).attr('id').replace('-bin', '');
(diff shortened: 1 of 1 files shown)


mboynes:
Thanks! I'll check this out and get it merged. In the meantime, you should check out https://github.com/alleyinteractive/fm-zones. It's in early development but I think it will ultimately replace DraggablePost.

kshaner:
Good to know! I'll take a look.

PR #364, opened 2015-07-13
Add term presave value filter

firejdl:
Add filter to Fieldmanager_Datasource_Term::presave_alter_values(), which allows themes and other plugins to filter the term values before they are saved.

### DIFF
--- a/php/datasource/class-fieldmanager-datasource-term.php
+++ b/php/datasource/class-fieldmanager-datasource-term.php
@@ -148,6 +148,8 @@ class Fieldmanager_Datasource_Term extends Fieldmanager_Datasource {
 			$values = array( $values );
 		}
 
+		$values = apply_filters( 'fm_datasource_term_presave_alter_values', $values, $field, $current_values, $this );
+
 		// maybe we can create terms here.
 		if ( get_class( $field ) == 'Fieldmanager_Autocomplete' && !$field->exact_match && isset( $this->taxonomy ) ) {
 			foreach( $values as $i => $value ) {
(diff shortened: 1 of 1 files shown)


dlh01:
Hey @firejdl, 

What's your use case for this filter?

firejdl:
We need primary tags to be specified in a separate field from regular tags, but we also want primary tags to be added as regular tags. This change allows for that.

dlh01:
Where is the primary tag stored?

firejdl:
It's stored as postmeta as well as one of the regular tags.

dlh01:
It sounds like you might be able to extend `Fieldmanager_Datasource_Term` to accomplish what you need. You could use a custom `save_taxonomy()` method that appended the primary tag after the rest of the tags are set. Something like:

```php

class My_Datasource extends Fieldmanager_Datasource_Term {

	public function save_taxonomy( $tax_values, $data_id ) {
		parent::save_taxonomy( $tax_values, $data_id );
		if ( $primary_tag = get_post_meta( $data_id, 'my_meta_key', true ) ) {
			wp_set_object_terms( $data_id, (int) $primary_tag, 'post_tag', true );
		}
	}

}
```

Would that work?

firejdl:
For one thing, we want to delete the old primary tag from the list of regular tags when you change primary tags. That's only possible with knowing the previous/current value, as in `presave_alter_values` (and the filter).

I realize that I could also extend `Fieldmanager_Datasource_Term`, modify the `$values` in `presave_alter_values()`, and then call `parent::presave_alter_values()`, but using a filter feels more like the WordPress way to do it...

firejdl:
We switched over to the extending `Fieldmanager_Datasource_Term` method a while back and I forgot to close this PR. :)

Issue #365, opened 2015-07-17
New repeatable datepickers initially fail to save their `pm` values


zgreen:
Newly added repeatable datepickers will fail to save their `pm` value, if selected, when `'use_time' => true` is set. This only happens on the initial post save, and only affects new groups or fields that are added, not those that already exist.

Steps to reproduce:
1. Using a repeatable datepicker field on the post edit screen, add a new group or field.
2. Add values to that field, and select `P.M.` from the select menu.
3. Save or publish the post.

The value will then be saved as A.M. If the field is updated to `P.M.`, the value will be saved properly.

Example markup of a new fm datepicker field:

```php
<div class="fm-item fm-test_date_picker_single fm-element"><div class="fmjs-removable-label fmjs-removable"><div class="fmjs-removable-element"><div class="fm-label fm-label-test_date_picker_single"><label for="fm-test_date_picker_single-proto">Start Date/Time</label></div></div><a href="#" class="fmjs-remove" title="Remove">Remove</a></div><input class="fm-element fm-datepicker-popup hasDatepicker fm-has-date-picker" type="text" data-datepicker-opts="{&quot;showButtonPanel&quot;:true,&quot;showOtherMonths&quot;:true,&quot;selectOtherMonths&quot;:true,&quot;dateFormat&quot;:&quot;d M yy&quot;}" name="test_date_picker_single[1][date]" id="fm-edit-dynamic-0">

	<span class="fm-datepicker-time-wrapper">
		@
		<input class="fm-element fm-datepicker-time" type="text" value="" name="test_date_picker_single[1][hour]">
		:
		<input class="fm-element fm-datepicker-time" type="text" value="" name="test_date_picker_single[1][minute]">
					<select name="test_date_picker_single[proto][ampm]">
				<option value="am">A.M.</option>
				<option value="pm">P.M.</option>
			</select>
			</span>
</div>
```

Note the `test_date_picker_single[proto][ampm]` on the `name` attribute of the select menu. This should be `test_date_picker_single[1][ampm]` in this case.

[...]

PR #366, opened 2015-07-17
add fm-element class to am/pm select

zgreen:
Addresses https://github.com/alleyinteractive/wordpress-fieldmanager/issues/365


mboynes:
:+1: Thanks @zgreen!

PR #368, opened 2015-07-26
Allow submenus of pages with query args

### LINKED ISSUE #195: Submenu context doesn't work for pages with secondary query args
This is easier to explain in code than words.

This works (note the `tools.php`):

```php
fm_register_submenu_page( 'page_submenu', 'tools.php', 'Submenu' );
add_action( 'fm_submenu_page_submenu', function() {
	$fm = new Fieldmanager_TextField( array( 'name' => 'page_submenu' ) );
	$fm->activate_submenu_page();
} );
```

This doesn't (note the `edit.php?post_type=page`):

```php
fm_register_submenu_page( 'page_submenu', 'edit.php?post_type=page', 'Submenu' );
add_action( 'fm_submenu_page_submenu', function() {
	$fm = new Fieldmanager_TextField( array( 'name' => 'page_submenu' ) );
	$fm->activate_submenu_page();
} );
```

This issue only affects the new context actions. The original syntax does work, e.g.:

```php
add_action( 'init', function() {
	$fm = new Fieldmanager_TextField( array( 'name' => 'page_submenu' ) );
	$fm->add_submenu_page( 'edit.php?post_type=page', 'Submenu' );
} );
```

This one might be tricky to resolve. I gave it the old college try and may have something workable. I wanted to log it separately as an issue in the event the PR won't cover it.

mboynes:
**Important**: My concern about this PR is that I'm moving the FM context hooks from `init` to `admin_init` when `is_admin()`. This is to allow the admin codebase to load so we can use core functions to calculate the current page (`get_plugin_page_hook()`). This could cause an issue if someone is executing code between init:99 and admin_init:5 and expecting their Fieldmanager fields to be defined.

Unfortunately, I couldn't figure out a reasonable way to test this using phpunit. It depends on numerous admin hooks firing and scripts loading (e.g. menu.php loading, and initializing the admin menu), which simply doesn't happen when running phpunit.

Resolves #195.

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -278,6 +278,8 @@ function fm_get_context() {
  * }
  */
 function fm_calculate_context() {
+	global $plugin_page, $typenow, $pagenow;
+
 	// Safe to use at any point in the load process, and better than URL matching.
 	if ( is_admin() ) {
 		$script = substr( $_SERVER['PHP_SELF'], strrpos( $_SERVER['PHP_SELF'], '/' ) + 1 );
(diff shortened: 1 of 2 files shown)


mboynes:
I don't like moving the context hooks outside of `init`, so I'm closing this one in favor of #369.

PR #370, opened 2015-07-29
Update prepare_data() to better use the passed field

mboynes:


bcampeau:
:horse:

php/context/class-fieldmanager-context.php
```diff
@@ -59,11 +59,11 @@ protected function prepare_data( $old_value = null, $new_value = null, $fm = nul
 			$fm = $this->fm;
 		}
 		if ( null === $new_value ) {
-			$new_value = isset( $_POST[ $this->fm->name ] ) ? $_POST[ $this->fm->name ] : '';
```
danielbachhuber: Let's revert this change for now, because we aren't fully certain it doesn't have unintended consequences.

PR #373, opened 2015-08-17
Change the action where meta boxes are removed

mboynes:
Many core meta boxes are not yet defined when `admin_init` runs; most are added immediately prior to `add_meta_boxes`, which is specifically documented as, "Fires after all built-in meta boxes have been added."


dlh01:
Is it worth using a late priority to be able to remove other meta boxes added on that hook? Other than that question, :deciduous_tree: from me.

mboynes:
@david Good thinking. I bumped it to 100.

PR #374, opened 2015-08-17
Improve flexibility of context actions

mboynes:


fieldmanager.php
```diff
@@ -418,17 +418,17 @@ function fm_trigger_context_action() {
...
+		do_action( "fm_{$context}_{$type}", $type );
 	}
+
+	/**
+	 * Fires when a specific Fieldmanager context loads.
+	 *
+	 * The dynamic portion of the hook name, $context, refers to the first
+	 * value returned by fm_calculate_context(). For example, the Edit User
+	 * screen would fire "fm_user".
+	 */
```
dlh01: I would note that the new `@param`s are missing docs.

mboynes: Good point, thanks!

fieldmanager.php
```diff
@@ -405,30 +405,36 @@ function fm_match_context( $context, $type = null ) {
...
-		 */
-		do_action( "fm_{$context}_{$type}" );
-	} else {
-		/**
-		 * Fires when a specific Fieldmanager context, but not type, loads.
 		 *
-		 * The dynamic portion of the hook name, $context, refers to the first
-		 * value returned by fm_calculate_context(). For example, the Edit User
-		 * screen would fire "fm_user".
+		 * @param string $type The context type, e.g. post, term, submenu.
```
dlh01: Wouldn't type be the post type name, taxonomy name, etc.?

dlh01:
One question, although I might be misreading the change. Otherwise, :deciduous_tree:

fieldmanager.php
```diff
@@ -243,12 +243,15 @@ function _fieldmanager_registry( $var, $val = null ) {
  *
  * @see fm_calculate_context() for detail about the returned array values.
  *
+ * @param bool $force Optional. If true, FM will recalculate the current context.
+ *                    This is necessary for testing and perhaps other
+ *                    programmatic purposes.
  * @return array Contextual information for the current request.
  */
-function fm_get_context() {
+function fm_get_context( $force = false ) {
```
dlh01: To me, `$force` reads as if you're forcing FM to return something when it otherwise wouldn't, which isn't quite what's happening, if I'm following correctly. Maybe FM could do with a dedicated `clear` or `recalculate` workflow, or a way to get a new calculated context independently of the static one? (I realize this comment is coming a little late in the 1.0 cycle, though.)

dlh01:
Looks :deciduous_tree: to me.

fieldmanager.php
mboynes: Good feedback. I think in the short term, I'll just rename this to `$recalculate`.

Issue #353, opened 2015-06-25
Not all FM scripts declare FM dependencies


dlh01:
In some cases, `fieldmanager.js` needs to be enqueued later than it normally is. 

One example is the Customizer (#331), where `fieldmanager.js` needs to follow some of core's scripts.

It's possible to change where `fieldmanager_enqueue_scripts()` is hooked, but not all of the field-specific scripts declare `fieldmanager_script` as a dependency. So `fieldmanager.js` will appear in the right spot, but other scripts can appear too early.

Adding dependencies to `fm_add_script()` calls might be enough to address this issue, but I'm sure there are other approaches, too.

dlh01:
Should have been taken care of in #376.

Issue #331, opened 2015-05-08
Customizer context


bcampeau:
We should consider adding a Customizer context. It would support all Fieldmanager fields in the left column and would also support a template to load in the right column for preview purposes.

There is heavy development on the Customizer for 4.3, so this might be worth waiting to fully consider until after that release.

mboynes:
This would be awesome. I think it's much trickier than it sounds, especially because the customizer has its own fields API. @dlh01 has done some one-off work on using FM with Customizer, he might be able to offer an informed take on it.

danielbachhuber:
> There is heavy development on the Customizer for 4.3, so this might be worth waiting to fully consider until after that release.

If you want to package as a standalone plugin for the time being, we would be happy beta testers :)

dlh01:
All I did in that one-off work was borrow the data in from some Fieldmanager objects to help construct Customizer objects. The data could have just as easily come from an array or something.

I'm starting a project in a couple weeks where I'd like to be able to try out an actual Customizer context. I'm interested in seeing what's possible.

dlh01:
The project I mentioned above did end up creating something of a Customizer context. Once #380 goes in, I plan to start working on a proper context here.

PR #376, opened 2015-08-28
Add `fieldmanager_script` as dependency where needed

dlh01:
See #353, #331. Adds `fieldmanager_script` as a dependency to the Autocomplete, Datepicker, and RichTextArea scripts. 

As far as I was able to tell, the rest of the field-specific scripts don't strictly depend on `fieldmanager_script`, so I didn't add the dependency to them. That said, there could be a future-proofing case for doing so.

Also adds some tests for script registration and dependencies.


netaustin:
:crown:

PR #378, opened 2015-09-11
Update Travis: Infrastructure, WordPress core, PHP

dlh01:


.travis.yml
```diff
@@ -8,12 +10,23 @@ php:
...
+    - WP_VERSION=4.1 WP_MULTISITE=1
+
+matrix:
+    include:
+        - php: 5.6
+          env: WP_VERSION=master
+        - php: 7.0
+          env: WP_VERSION=master
+    allow_failures:
+        - php: 5.6
```
mboynes: Let's not allow 5.6 to fail

.travis.yml
```diff
@@ -8,12 +10,23 @@ php:
...
-    - WP_VERSION=3.9 WP_MULTISITE=0
-    - WP_VERSION=3.9 WP_MULTISITE=1
-    - WP_VERSION=3.8 WP_MULTISITE=0
-    - WP_VERSION=3.8 WP_MULTISITE=1
+    - WP_VERSION=4.3 WP_MULTISITE=0
+    - WP_VERSION=4.3 WP_MULTISITE=1
+    - WP_VERSION=4.2 WP_MULTISITE=0
+    - WP_VERSION=4.2 WP_MULTISITE=1
+    - WP_VERSION=4.1 WP_MULTISITE=0
+    - WP_VERSION=4.1 WP_MULTISITE=1
```
mboynes: Do you think it's even worth doing 4.1? I think I'd rather remove that, restrict it to 3 versions, and add php 5.6 to the full php version array

mboynes:
Bounced some ideas, but this is good to go regardless :+1:

.travis.yml
dlh01: I don't have an opinion about 4.1. I was only following the previous pattern, and I'm not sure what the original reasoning was for that. I'll just drop it and move 5.6.

PR #379, opened 2015-09-14
Selecting media attachment when opening frame

srtfisher:


bcampeau:
@mboynes should review this too

bcampeau:
It looks OK to me, but please hold on merging it

js/media/fieldmanager-media.js
```diff
@@ -57,7 +57,23 @@ $( document ).on( 'click', '.fm-media-button', function( event ) {
 		$wrapper.html( preview ).trigger( 'fieldmanager_media_preview', [ $wrapper, attachment, wp ] );
 	});
 
+	// Select the attachment when the frame opens
+	fm_media_frame[ $el.attr('id') ].on( 'open', function() {
+		// Select the current attachment inside the frame
+		var selection = fm_media_frame[ $el.attr('id') ].state().get('selection'),
+			id = $el.parent().find('.fm-media-id').val(),
```
mboynes: just want to confirm, there's no way there would be more than one `.fm-media-id` in `$el`'s parent, right?

srtfisher: `$el` is a reference to that specific element's button (be that the media select button or the specific media select button for that attachment in a group. The parent would be `.fm-item`, either the nested `.fm-item` for the item in the group OR the general `.fm-item` for that field. Since we know that, we are able to properly scope the `.fm-media-id` hidden input.

We're also using [similar logic](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/js/media/fieldmanager-media.js#L46) on the 'select' callback.

mboynes:
:100:. I tried _really_ hard to break this and I failed miserably. Nice work!

PR #380, opened 2015-09-15
Add QUnit tests

dlh01:
This would add a basic infrastructure for adding and running JS unit tests with QUnit. 

It could use a good look from anyone with some QUnit experience. I'm sure a lot of it could be improved, now or later, such as relying on the `develop` repository and the "script loader" in `index.html`.


dmachat:
lgtm! :+1:

Issue #387, opened 2015-10-08
Fieldmanager_Media fields don't allow preview size to be set.


goldenapples:
The [docs](http://api.fieldmanager.org/source-class-Fieldmanager_Media.html#39-45) show that it should be possible to set the preview size of a Fieldmanager_Media field to any registered image size with syntax like:

```
'featured_image' => new \Fieldmanager_Media( esc_html__( 'Featured Image' ), array(
    'name'            => 'featured_image',
    'preview_size'    => 'medium',
    'thumbnail_class' => 'variant-featured-image-preview',
    'description'     => esc_html__( 'Alternate featured media header for this variant.' ),
    ) ),
```

However, this doesn't work when initially adding the image.

The issue seems to be the inlined javascript [here](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/class-fieldmanager-media.php#L119) which defines the `fm_preview_sizes` variable as an array rather than an object. (As such, trying to set or get named properties of the object will always fail.) Changing that line to `var fm_preview_size = fm_preview_size || {};` causes media selection to respect the size chosen on initial preview.

mboynes:
This should be fixed as of #409, but please do re-open if you're still experiencing it!

PR #388, opened 2015-10-08
Define `fm_preview_sizes` var as object, not array

goldenapples:
Since the properties on it are set as named properties, not by numeric
indexes, this variable needs to be defined as an object, or attempts to
set or read it's properties will fail.

See #387

### DIFF
--- a/php/class-fieldmanager-media.php
+++ b/php/class-fieldmanager-media.php
@@ -116,7 +116,7 @@ class Fieldmanager_Media extends Fieldmanager_Field {
 			<input type="hidden" name="%2$s" value="%4$s" class="fm-element fm-media-id" />
 			<div class="media-wrapper">%5$s</div>
 			<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


dlh01:
If memory serves, I've hit this issue before as well. Of course, I can't seem to replicate it now, even with your example code in #387. But I might not have the right test set up, though. 

Either way, this seems OK to me, but I'll leave it open for now to get a second opinion or the merge.

goldenapples:
Thanks. 

For reproduction steps, I'm using Firefox Mac and uploading a very large image, for which the preview shows up full size, rather than the image size specified:

screen shot 2015-09-03 at 3 01 11 pm

screen shot 2015-09-03 at 3 10 45 pm

dlh01:
@goldenapples Is this still an issue after #409?

mboynes:
Yeah, #409 should have fixed this. Good catch and thanks for the PR all the same!

PR #393, opened 2015-10-23
Dynamically set `Fieldmanager_field::$is_attachment`

danielbachhuber:
If we're adding the meta box to an attachment page, we can safely assume the `save_fields_for_attachment()` method needs to be called.


danielbachhuber:
@mboynes #reviewmerge

PR #396, opened 2015-11-04
Select field required/first_empty

jasonbahl:
Allow select fields to be "required" and set with "first_empty"


jasonbahl:
Closing this due to inactivity for over 3 years now. . .

PR #399, opened 2015-11-15
Add Customizer context

### LINKED ISSUE #331: Customizer context

dlh01:
This is a work-in-progress PR to add a Customizer context. If you want to take it for a spin, try the companion branch in Fieldmanager Demos: https://github.com/alleyinteractive/fieldmanager-demos/blob/customizer-context/lib/class-fm-demo-context-customizer.php

What's included so far covers most of Fieldmanager's fields and their features, but questions about the implementation and known bugs are below.

Suggestions and feedback are appreciated!

## Bugs

- [x] `Fieldmanager_Datepicker` doesn't work.
- [x] `Fieldmanager_RichTextArea` doesn't work (and I'm not sure what would go into fixing this).
- [x] Unnecessary `_debounce()` applies when updating an `Autocomplete`'s hidden field.
- [x] There's no signal to the user if a field fails validation.
- [x] Any changes to field values from sanitizing aren't reflected in the form.
- [x] JS callbacks are not bound to dynamically added sections.
- [x] `Fieldmanager_Colorpicker` doesn't work.
- [x] Inline docs about the architecture are insufficient.
- [x] Missing unit tests.

## Questions

* Is there a better way to support `postMessage` settings? Currently, these settings work, but the preview receives a serialized string, and none of Fieldmanager's processing (e.g., validation) occurs beforehand. Replacing jQuery's `serialize()` with serializing into JSON (which would then be passed to both JS and PHP) could help here.
* Is there a better way to reinitialize FM sortable fields or label macros than adding event handlers to `fieldmanager.js`?
* Is it a problem to manually set the `Autocomplete`'s hidden field value to "", which doesn't happen in other contexts?

Resolves #331.


danielbachhuber:
:100:

dlh01:
The last round of changes to this PR should have fixed `Datepicker` fields and the `_debounce()` delay on `Autocomplete` fields. It also fixed `display_if`, which wasn't working at all.

> Is there a better way to reinitialize FM sortable fields or label macros than adding event handlers to fieldmanager.js?

I've answered this question for now as "yes" and added a `fm_customizer_control_section_expanded` trigger, which other scripts in FM can bind to. This seems a little more consistent to me with other FM JS behavior.

dlh01:
> Any changes to field values from sanitizing aren't reflected in the form.

This seems to be an issue with core controls as well, so I've marked it as "done," for now, in that Fieldmanager doesn't introduce it.

js/fieldmanager-autocomplete.js
```diff
@@ -77,5 +77,9 @@ fm.autocomplete = {
 
 $( document ).ready( fm.autocomplete.enable_autocomplete );
 $( document ).on( 'fm_collapsible_toggle fm_added_element fm_displayif_toggle fm_activate_tab', fm.autocomplete.enable_autocomplete );
+$( document ).on( 'focus',
+	'input[class*="fm-autocomplete"]:not(.fm-autocomplete-enabled)',
```
danielbachhuber: Do we need the prior to implementation approaches in addition to this approach?

js/fieldmanager.js
```diff
@@ -114,6 +114,25 @@ var match_value = function( values, match_string ) {
 	return false;
 }
 
+/**
+ * Initializes triggers to conditionally hide or show fields.
+ */
+var init_display_if = function() {
+	$( '.display-if' ).each(function() {
```
danielbachhuber: It seems like this could collide with another class in global scope. Can we narrow scope?

dlh01: I agree, although because this is existing FM code just relocated, I'd be inclined to determine a better scope in a separate PR. If that sounds OK to you, I'll file an issue.

js/fieldmanager-autocomplete.js
dlh01: I don't think so. But the existing JS predates my involvement with FM, so I'd appreciate a review from you or someone else to confirm.

js/fieldmanager.js
danielbachhuber: > If that sounds OK to you, I'll file an issue.

Sounds fine

dlh01:
0795c83 updates the behavior in `Fieldmanager_Field::_failed_validation()` to account for the Customizer, and it provides an `alert()` to the user if validation fails while attempting to save changes.

At the moment, there doesn't seem to be a standard way to alert users to errors while saving changes in the Customizer. `alert()` is at least consistent with the "Are you sure" prompt, but it still isn't great, and suggestions for alternate approaches are welcome.

There also doesn't seem to be a great a way to send a message when an error occurs while previewing a setting change. Failing validation will still stop the preview from using invalid values, but it does so silently. Suggestions welcome here, too.

dlh01:
> Is there a better way to support postMessage settings?

01893e7 adds the [serializeJSON](https://github.com/marioizquierdo/jquery.serializeJSON) library for converting the value of an FM field inside a control into an object, which can then be handled more easily in PHP or JS.

dlh01:
8066242 and the few commits preceding it should get `Fieldmanager_RichTextArea`s working in the Customizer context.

dlh01:
>  Inline docs about the architecture are insufficient.

I'm happy to add more documentation where needed, but the explanation I had in mind when I added that todo is in 8bb2f5d.

dlh01:
Closing for now to reconsider a few things about the context.

dlh01:
Reopening, now with 2396297, which attempts to increase the context's flexibility by loosening its connection to Customizer sections.

dlh01:
The proposed Customize context is now available as a standalone plugin that you can use alongside a stable version of Fieldmanager: [Check out Fieldmanager Beta: Customize](https://github.com/dlh01/fieldmanager-beta-customize) (props @jameswburke for reviewing the initial pull request).

Screenshots, installation instructions, details about how the plugin integrates with Fieldmanager, and steps to activate a demo are all available in [the README](https://github.com/dlh01/fieldmanager-beta-customize/blob/master/README.md).

If you're interested in using the Customize context, please try the plugin. Feedback, bug reports, and pull requests are all welcome at the repo. 

Plugin development will continue much like a feature plugin in WordPress core. Improvements to the plugin based on testing will, hopefully, strengthen this pull request in turn.

PR #401, opened 2015-11-16
Add support for core term meta

### LINKED ISSUE #400: Add support for core's term meta and deprecate FM term meta
As of WordPress 4.4, term meta is built into core. Fieldmanager has always provided an avenue for term meta by creating hidden posts 1:1 with terms, and using post meta to store the data. Fieldmanager's term meta should be deprecated, and support should be added for core's term meta.

This is a delicate situation, as we don't want to break backwards compatibility, but we also don't want the default to be a deprecated (and highly discouraged) option. In order to accomplish this, we're going to deprecate the current helper method `Fieldmanager_Field::add_term_form()` in favor of a new method `Fieldmanager_Field::add_term_meta_box()`, and we're going to deprecate the current argument order for `Fieldmanager_Context_Term::__construct()` to prefer arguments passed as an array instead of being passed separately. The default for the deprecated method and construct arguments will be to use Fieldmanager's term meta, and the default for the new method and construct argument array will be to use core's term meta.

* [x] Deprecate `Fieldmanager_Field::add_term_form()`, ensuring that the default term meta storage remains Fieldmanager's term meta.
* [x] Add `Fieldmanager_Field::add_term_meta_box()` which sets the default term meta storage to be core's.
* [x] Deprecate the individual arguments passed to `Fieldmanager_Context_Term::__construct()`, ensuring that the default term meta storage remains Fieldmanager's term meta.
* [x] Add support for arguments passed to `Fieldmanager_Context_Term::__construct()` as an array. The default term meta storage should be core's when creating the object this way (however, it should be able to be overridden).
* [x] Ensure that existing test coverage for Fieldmanager term meta remain intact.
* [ ] Add full test coverage for using core term meta.
* [x] Build a migration tool to move all Fieldmanager term meta to core term meta. This should be a separate plugin.

mboynes:
Resolves #400


bcampeau:
:horse:

php/class-fieldmanager-field.php
```diff
@@ -979,8 +983,49 @@ public function add_page_form( $uniqid ) {
 	 * @param int $parent Only show this field on child terms of this parent term ID
 	 */
 	public function add_term_form( $title, $taxonomies, $show_on_add = true, $show_on_edit = true, $parent = '' ) {
+		_deprecated_function( __METHOD__, 'Fieldmanager-1.0.0-beta.3', 'Fieldmanager_Field::add_term_meta_box()' );
```
danielbachhuber: Do we need a WordPress version check here?

If I'm running < WordPress 4.4, I think I'd need to continue to use `add_term_form()`

danielbachhuber: Also, see recent `wp_title()` [debacle](https://core.trac.wordpress.org/changeset/35624) to see the impact of calling `_deprecated_function()` too soon.

I think `add_term_meta_box()` should exist for a couple to few WordPress releases before you formally deprecate `add_term_form()`. Otherwise, developers will need to have a lot of conditionals in their code if they're working against trunk locally, and using an outdated version of WordPress in production (cough cough VIP)

php/context/class-fieldmanager-context-term.php
```diff
@@ -58,29 +72,75 @@ class Fieldmanager_Context_Term extends Fieldmanager_Context_Storable {
...
+			$this->taxonomies   = (array) $args['taxonomies'];
+			$this->show_on_add  = $args['show_on_add'];
+			$this->show_on_edit = $args['show_on_edit'];
+			$this->parent       = $args['parent'];
+			$this->use_fm_meta  = $args['use_fm_meta'];
+			$this->fm           = $args['field'];
+		} elseif ( empty( $taxonomies ) ) {
+			throw new FM_Developer_Exception( esc_html__( '"title" and "taxonomies" are required values for Fieldmanager_Context_Term', 'fieldmanager' ) );
+		} else {
+			_deprecated_argument( __METHOD__, 'Fieldmanager-1.0.0-beta.3', esc_html__( 'Instantiating Fieldmanager_Context_Term using individual arguments is deprecated. Please pass an array of arguments instead.', 'fieldmanager' ) );
```
danielbachhuber: Again, I think it's fine to support two approaches for passing arguments, but I don't think it makes sense to immediately start calling `_deprecated_*()`

tests/php/test-fieldmanager-context-term.php
```diff
@@ -110,9 +122,13 @@ private function _get_html_for( $field, $test_data = null ) {
 	}
 
 	public function test_context_render_add_form() {
+		if ( ! _fm_phpunit_is_wp_at_least( 4.4 ) ) {
+			return $this->_skip_tests_because_version( 4.4 );
```
danielbachhuber: Why not test the Fieldmanager interfaces regardless of WordPress version?

I'd think you'd want to test that term meta works for both WordPress 4.3 and WordPress 4.4

danielbachhuber:
Looks pretty good, aside from the (imo) over-eager deprecation of existing APIs.

fieldmanager.php
```diff
@@ -3,7 +3,7 @@
  * Fieldmanager Base Plugin File.
  *
  * @package Fieldmanager
- * @version 1.0.0-beta.2
+ * @version 1.0.0-beta.3
```
mboynes: Ideally yes -- I included it here because of the deprecations (even if only in phpdoc).

php/class-fieldmanager-field.php
mboynes: These are great points; thanks for the input!

tests/php/test-fieldmanager-context-term.php
mboynes: There's another test class for FM term meta, where everything here is also tested using the old method. I'm skipping these tests as a convenience -- otherwise, I'd have to do a level of error-checking that just seems unnecessary, e.g. I couldn't call `$base->add_term_meta_box( 'test meta box', $this->taxonomy )->add_term_fields( $this->taxonomy );`, I'd have to:
```php
$context = $base->add_term_meta_box( 'test meta box', $this->taxonomy );
if ( $context ) {
    $context->add_term_fields( $this->taxonomy );
}
```
... then conditionally run the assertions.

PR #402, opened 2015-11-16
Bug fixes for term datasource

### LINKED ISSUE #386: Unit tests fail after r34737
As Travis has noted, Fieldmanager's unit tests are failing after [r34737](https://core.trac.wordpress.org/changeset/34737) in core -- specifically, [`test_datasource_term_save_empty()`](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/tests/php/test-fieldmanager-datasource-term.php#L194), `Failed asserting that null is identical to ''`.

It looks like we need to investigate whether this change poses a larger problem for how FM manages data than updating the types in that one test.

mboynes:
* Fix a data type issue introduced in core, because `null !== ''` (see #386 for further details; this resolves #386)
* Ensure any taxonomy used in a datasource, where the data is stored as a term relationship, is sortable

### DIFF
--- a/php/datasource/class-fieldmanager-datasource-term.php
+++ b/php/datasource/class-fieldmanager-datasource-term.php
@@ -76,15 +76,20 @@ class Fieldmanager_Datasource_Term extends Fieldmanager_Datasource {
 		}
 
 		parent::__construct( $options );
-		if ( $this->only_save_to_taxonomy ) $this->taxonomy_save_to_terms = True;
 
-		// make post_tag and category sortable via term_order, if they're set as taxonomies, and if
-		// we're not using Fieldmanager storage
-		if ( $this->only_save_to_taxonomy && in_array( 'post_tag', $this->get_taxonomies() ) ) {
-			$wp_taxonomies['post_tag']->sort = True;
+		// Ensure that $taxonomy_save_to_terms is true if it needs to be
+		if ( $this->only_save_to_taxonomy ) {
+			$this->taxonomy_save_to_terms = true;
 		}
-		if ( $this->only_save_to_taxonomy && in_array( 'category', $this->get_taxonomies() ) ) {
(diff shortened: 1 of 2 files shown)


php/datasource/class-fieldmanager-datasource-term.php
```diff
@@ -76,15 +76,20 @@ public function __construct( $options = array() ) {
...
+		// Ensure that $taxonomy_save_to_terms is true if it needs to be
+		if ( $this->only_save_to_taxonomy ) {
+			$this->taxonomy_save_to_terms = true;
 		}
-		if ( $this->only_save_to_taxonomy && in_array( 'category', $this->get_taxonomies() ) ) {
-			$wp_taxonomies['category']->sort = True;
+
+		if ( $this->taxonomy_save_to_terms ) {
+			// Ensure that the taxonomies are sortable if we're not using FM
+			// storage
```
dlh01: Odd-looking widow on this line.

Also, core standards suggest sentences in docs are supposed to end with...eh, never mind.

php/datasource/class-fieldmanager-datasource-term.php
```diff
@@ -76,15 +76,20 @@ public function __construct( $options = array() ) {
...
+			foreach ( $this->get_taxonomies() as $taxonomy ) {
+				if ( ! empty( $wp_taxonomies[ $taxonomy ] ) ) {
```
dlh01: Although they're functionally the same thing, would it be preferable to use `taxonomy_exists()`?

dlh01:
Two minor comments, otherwise looks :deciduous_tree: to me.

php/datasource/class-fieldmanager-datasource-term.php
mboynes: We're already using `$wp_taxonomies`, so we may as well save the function call IMO

PR #403, opened 2015-11-17
Allow users to restrict media type on Fieldmanager_Media

firejdl:
This commit adds a "media_type" parameter on the Fieldmanager_Media object, which allows developers to restrict the media type for the Media Library that is opened. Valid options are "all", "image", "audio", or "video".

### DIFF
--- a/js/media/fieldmanager-media.js
+++ b/js/media/fieldmanager-media.js
@@ -22,10 +22,10 @@ $( document ).on( 'click', '.fm-media-button', function( event ) {
 		return;
 	}
 
-	// If media type has been restricted, make sure the library only shows that
+	// If mime type has been restricted, make sure the library only shows that
 	// type.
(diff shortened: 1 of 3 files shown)


mboynes:
Turns out this PR is more capable than you realize. It looks like you can use any mime type like "application/pdf", or the first part of any mime type (before the slash), like "application". That's awesome.

Can you...
* [ ] Update the inline docs to cover this
* [ ] Merge in the latest master (that should fix the failing unit tests)
* [ ] Change `"media_type"` to `"mime_type"`
* [ ] Add a basic unit test (just verify that when adding `mime_type` that the `data-type` attribute is successfully added)

firejdl:
Hi @mboynes:

I updated the pull request with the changes you requested, and also fixed a bug that we found when the user can add multiple items (`$limit` > 1).

mboynes:
Replaced by #416

PR #409, opened 2015-12-10
Fix repeatable media field preview sizes

mboynes:
This fixes a bug where repeatable media fields would show the wrong preview size for newly-created fields. Also, adds basic unit tests for rendering and saving.


bcampeau:
:horse:

PR #417, opened 2015-12-23
Make `Fieldmanager_Options::$sanitize` a non-closure function

firejdl:
`Fieldmanager_Options::$sanitize` was defined as a closure, which means that objects of this class are not serializable. This commit makes `$sanitize` a regular function.

This is similar to #357, and also alleyinteractive/fm-zones#3


mboynes:
:+1: Thanks!

PR #421, opened 2015-12-31
Fix quickedit unit tests

mboynes:
Travis will still fail due to other issues, but this needs to be merged before fixing those.


romancandlethoughts:
:rocket:

PR #422, opened 2015-12-31
Fix unit tests for multi tool labels

mboynes:
These broke in #419, but the bug in #421 prevented Travis from reporting it.

Note that Travis will still fail on the latest nightly builds of core, due to a recent core update surrounding richtextareas.


romancandlethoughts:
:computer:

PR #428, opened 2016-01-06
Add wildcards to get_users search

### LINKED ISSUE #47: Autocomplete on users needs enhancement
Autocomplete on users currently uses the [`get_users()`](http://codex.wordpress.org/Function_Reference/get_users) "search" argument. This is quite limited for two reasons:

1. You can't search by display name, which is not only the most intuitive way to search for a user, but it's also what the autocomplete returns. From the Codex, "Use this argument to search users by email address, URL, ID or username (this does not currently include display name)." 
2. The searches must be exact, which defeats the purpose of autocomplete. That is, if you're searching for the user "John Smith", whose username is "admin", you would have to type "admin". Typing "admi" would get you nothing. And per the first point, once you type "admin", the autocomplete will say "John Smith".

We should implement a custom search solution for this to resolve both issues.

mboynes:
Closes #47.

### DIFF
--- a/php/datasource/class-fieldmanager-datasource-user.php
+++ b/php/datasource/class-fieldmanager-datasource-user.php
@@ -141,7 +141,7 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
         $ret = array();
 
         if ( $fragment ) {
-        	$user_args['search'] = $fragment;
+        	$user_args['search'] = '*' . $fragment . '*';
         }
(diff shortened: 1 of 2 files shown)


enrico-sorcinelli:
Hi,
thanks for the merge and sorry if I didn't updated unit tests too.

   \- Regards

mboynes:
No problem at all, thanks again for the PR!

PR #429, opened 2016-01-07
Autocomplete highlight search character sequence in drop-down results

enrico-sorcinelli:
This patch allows to highlight the search character sequence in drop-down results in autocomplete fields. For example, if I have an item: "foo bar baz" it and I search for "ba" I'll see "foo **ba**r **ba**z" in drop down.

The patch extends jQuery UI autocompleter widget through the widget factory pattern.
In order to do the highlight, a `span` with `fm-autocomplete-highlight` class is wrapped around the sequence.

The highlight can be turned on/off with `highlight` constructor option (default to `true`), i.e.:

```php
    'highlight' => true,
    'name' => 'size',
    'datasource'  => new Fieldmanager_Datasource( array(
        'options' => array( 'Small', 'Medium', 'Large' ),
    ) )
) );
```

Feel free to change CSS class name, styles and option name I used.

PS: I searched over the repository and it seems that a similar functionality or pull request doesn't yet exists. 
Actually, there is a component that does a similar thing (_docs-src/templates/bootstrap/js/jquery.autocomplete.js_) but it seems that is related only to the documentation part.

### DIFF
--- a/css/fieldmanager.css
+++ b/css/fieldmanager.css
@@ -370,3 +370,9 @@ a.fm-delete:hover {
 .form-field .fm-checkbox label {
 	display: inline;
 }
+
+/* Autocomplete */
+.fm-autocomplete-highlight {
+	font-weight: bold;
+	color: #cc0000;
+}
--- a/js/fieldmanager-autocomplete.js
+++ b/js/fieldmanager-autocomplete.js
@@ -68,12 +68,44 @@ fm.autocomplete = {
 					} );
 				}
 
+				// Set autocomplete highligth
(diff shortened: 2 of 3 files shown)


enrico-sorcinelli:
Let me know if there's anything else I can/need to do to get the PR merged in (for example unit tests).

It must be said that you can obtain the same effect by adding the JS and CSS parts on a separate files (as in fact I'm currently doing in my Fieldmanager installations); the server part allows only to turn on/off the highlighting per-field. 
Of course having it in the plugin would be more comfortable

PR #430, opened 2016-01-07
Hopefully cut travis' workload in half

mboynes:


danielbachhuber:
What does it do?

mboynes:
[Source](http://stackoverflow.com/questions/31882306/how-to-configure-travis-ci-to-build-pull-requests-merges-to-master-w-o-redunda).

Every commit to every PR is being run twice because we have "Build pushes" and "Build pull requests" both enabled in Travis. This will stop Travis from checking a push if the push is not against `master`. You can see this PR only had one check, whereas #418, for example, has two.

danielbachhuber:
Oh, huh. That's awesome — and insane default behavior for Travis.

greg-1-anderson:
#drush had this problem too. https://github.com/drush-ops/drush/pull/1906

mbijon:
@danielbachubber Travis' defaults are based on "cheap" CI time vs "expensive" developer time:

It's good to be nice to Travis' system, like here.

But you don't want to only test against only 2 versions of PHP, for example. Then, a version-specific error leaking past poor CI coverage might create an expensive, rush dev bug-fix to a new version in Prod. Meanwhile an extra 5-10 minutes of machine time might have prevented the release and kept everything low-priority.

mboynes:
@mbijon While I agree that we should keep in mind that machine time is cheaper than dev time, when CI builds are run twice against the same commit (and therefore, builds take twice as long as they should), it affects dev time without added benefit. This update wasn't an attempt at being nice to Travis, it was to speed up the silly amount of time I had to wait to merge PRs.

mbijon:
You must have misunderstood me @mboynes. By "It's good to be nice to Travis' system, like here" i meant that this PR is a good idea. +1

I was addj g color b/c I've noticed a recent trend to not test on all active versions of PHP. This ticket does NOT toe that line, and it wasnt 100% needed for mw to make such a long comment.

PR #434, opened 2016-01-21
Recognize term.php as context for "term"

### LINKED ISSUE #433: Term context should support the new `term.php` page introduced in core
In the current trunk of core, the admin page for editing a term is `term.php` rather than `edit-tags.php`. This falls through the switch in the `fm_calculate_context()` function in Fieldmanager, and as a result term meta fields aren't being shown on the term edit page.

goldenapples:
Since https://core.trac.wordpress.org/changeset/36308, the link for editing a term is `term.php`, not `edit-tags.php?action=edit`. This looks for that screen context when calculating Fieldmanager context values.

Fixes #433.

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -355,7 +355,7 @@ function fm_calculate_context() {
 				break;
 			// Context = "term".
 			case 'edit-tags.php':
-            case 'term.php':
+			case 'term.php':
 				if ( !empty( $_POST['taxonomy'] ) ) {
 					$calculated_context = array( 'term', sanitize_text_field( $_POST['taxonomy'] ) );
 				} elseif ( !empty( $_GET['taxonomy'] ) ) {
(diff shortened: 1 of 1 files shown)


fieldmanager.php
```diff
@@ -355,6 +355,7 @@ function fm_calculate_context() {
 				break;
 			// Context = "term".
 			case 'edit-tags.php':
+            case 'term.php':
```
danielbachhuber: Indentation is off?

danielbachhuber:
I feel like the core commit might be reverted, given the changeset causes issues with backwards compat.

goldenapples:
Fair enough. This is a simple fix for running Fieldmanager against nightly core builds, but it probably doesn't need to be merged until/unless that changeset makes it into a release candidate.

danielbachhuber:
I'd be fine landing it. Could you add some code comment referencing the code ticket though, so we have history in the codebase for later?

goldenapples:
Sure, added in 357427f.

Issue #352, opened 2015-06-25
Repeatable Multi-select fields throw exception when empty


mboynes:
The following code will throw an exception if you don't select any options:

```
$fm = new Fieldmanager_Select( array(
	'name'    => 'multi_select',
	'options' => array( 'One', 'Two', 'Three' ),
	'limit' => 0,
	'multiple' => true,
) );
$fm->add_meta_box( 'Repeatable Multiselect', 'post' );
```

The exception is: `FM_Exception: $values should be an array because $limit is 0 in /var/www/wordpress/wp-content/plugins/wordpress-fieldmanager/php/class-fieldmanager-field.php on line 1048`.

PR #442, opened 2016-02-19
fix handling for repeatable multiple select fields #352

nschoenholtz:
I believe that this is an appropriate fix for #352, @mboynes please sanity check me?


mboynes:
:+1:

PR #444, opened 2016-02-19
accommodate fields with literal string value "0" and add test

### LINKED ISSUE #132: "0" (a string) is misconstrued as an empty value in textfields

xmattus:
Fixes #132


danielbachhuber:
:hammer:

PR #449, opened 2016-02-20
Define custom `sanitize` callback early, to allow override by `$options`

### LINKED ISSUE #287: `sanitize` can't be set when instantiating field that extends Fieldmanager_Options
In this field:

```php
$fm = new Fieldmanager_Select( 'Basic Select', array(
    'name'    => 'basic_select',
    'options' => array( 1, 2, 3 ),
    'sanitize' => 'absint',
) );
```

the `sanitize` argument will be overwritten in `Fieldmanager_Options::__construct()`: https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/class-fieldmanager-options.php#L68-L74.

`Fieldmanager_Radios` and `Fieldmanager_Checkboxes` are similarly affected.

danielbachhuber:
Fixes #287

### DIFF
--- a/php/class-fieldmanager-options.php
+++ b/php/class-fieldmanager-options.php
@@ -55,6 +55,7 @@ abstract class Fieldmanager_Options extends Fieldmanager_Field {
 	 * @param mixed $options
 	 */
 	public function __construct( $label = '', $options = array() ) {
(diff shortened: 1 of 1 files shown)


mboynes:
:+1:

PR #457, opened 2016-02-20
When an element is dynamically added, update its label's `for` too

### LINKED ISSUE #333: Labels "for" attributes aren't updated when created from prototype
Given the following code,

```php
$fm = new Fieldmanager_TextField( array(
	'name' => 'demo-field',
	'limit' => 0,
	'label' => 'Text Field',
) );
$fm->add_meta_box( 'Demo', 'post' );
```

When one clicks "Add Field", the generated HTML is:

```html
<div class="fm-item fm-demo-field fm-text">
  <div class="fmjs-removable-label fmjs-removable">
    <div class="fmjs-removable-element">
      <div class="fm-label fm-label-demo-field">
        <label for="fm-demo-field-proto">Text Field</label>
      </div>
    </div>
    <a href="#" class="fmjs-remove" title="Remove">Remove</a>
  </div>
  <input class="fm-element" type="text" name="demo-field[1]" id="fm-edit-dynamic-0" value="" size="50">
</div>
```

Note that the "id" attribute on the input field is correctly set to "fm-edit-dynamic-0" but the "for" attribute on the label remains "fm-demo-field-proto".

danielbachhuber:
Fixes #333

### DIFF
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -88,6 +88,10 @@ var fm_renumber = function( $wrappers ) {
 							$( this ).attr( 'name', new_fname );
 							if ( $( this ).attr( 'id' ) && $( this ).attr( 'id' ).match( '-proto' ) && ! new_fname.match( 'proto' ) ) {
 								$( this ).attr( 'id', 'fm-edit-dynamic-' + dynamic_seq );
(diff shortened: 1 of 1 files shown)


mboynes:
:+1:

PR #459, opened 2016-02-20
Address some buggy display-if cases

### LINKED ISSUE #241: Checkboxes should also support "display-if"

### LINKED ISSUE #242: Display-if doesn't support boolean values.
If i a use display-if with Drop Down list (having values true and false), i get JS errors and UI breaks. 
screen shot 2014-10-29 at 12 04 08 pm

### LINKED ISSUE #405: Autocomplete field does not work as a display_if src.
While trying to use an Autocomplete field as a display_if src, I found that because both the dummy and 'real' fields end up with the `.display-trigger` class, the dummy field causes a Javascript error on this line when it finds there is no `name` attribute after selecting a value from the autocomplete list:
https://github.com/alleyinteractive/wordpress-fieldmanager/blob/86deda5037a8350eaf796c170e51b3133ceb6f6e/js/fieldmanager.js#L187

Through a series of events I don't quite follow, this error also results in the value of the selected item not being saved to the 'real' field, and thus the change trigger never fires for that field and allowing the hidden field to show, and also results in the field value never being saved.

I could see either the dummy field not getting the `.display-trigger` class or adding a check to skip processing if the `name` attribute on a field doesn't exist, maybe even both just to be safe.

### LINKED ISSUE #435: display_if does not work with numeric values
On setting `value` to an integer or numeric-only string, I receive the following JS error when viewing the fields:

`Uncaught TypeError: $(...).data(...).split is not a function` in [fieldmanager.js L171 in my version](https://github.com/spiritedmedia/wordpress-fieldmanager/blob/master/js/fieldmanager.js#L171)

If I add non-numeric characters to the value, the error does not occur (but then of course I'm unable to check for the correct value).

It looks like #418 might address this in 170f09ecf9818fd0fdb46478e0fce353627e24a2.

In my specific case, I'm using the term datasource, so term IDs are stored in the field. If anyone has a workaround for this case, I'd appreciate it.

dlh01:
This combines effort and notes from several previous reports and PRs: #243, #415, #418, #435, #264, #324. It attempts to focus on bugs fixed in that work but leaves out some enhancements for post-1.0. 

- Allow numeric values to be used as the display-if 'value'
- Allow checkboxes to be used as display-if triggers
- Allow autocompletes to be used as display-if triggers
- Improve some cases of handling field visibility on load

Closes #243, fixes #242, closes #415, fixes #435, closes #264, fixes #405, fixes #241.

### DIFF
--- a/js/fieldmanager-autocomplete.js
+++ b/js/fieldmanager-autocomplete.js
@@ -34,7 +34,7 @@ fm.autocomplete = {
 								custom_data = custom_result;
 							}
 						}
-						
+
 						$.post( ajaxurl, {
 							action: $el.data( 'action' ),
 							fm_context: $el.data( 'context' ),
@@ -57,7 +57,14 @@ fm.autocomplete = {
 							$hidden.val( '' );
 							$el.val( '' );
 						}
-					}
+					};
+
+					// Handle the user deleting the input text, which is not an AC 'change'.
+					$( this ).on( 'keyup', function( e ) {
+						if ( e.target && '' === $( e.target ).val() ) {
+							$hidden.val( '' ).trigger( 'change' );
+						}
+					});
 				} else {
 					$( this ).on( 'keyup', function( e ) {
 						if ( e.keyCode == 27 || e.keyCode == 13 ) {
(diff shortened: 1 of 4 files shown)


mboynes:
:+1:

PR #462, opened 2016-02-20
remove php7 from allowed failures

aaronjorbin:
PHP7 is fully supported by WordPress core, it should be fully supported by fieldmanager


mboynes:
:+1:

PR #466, opened 2016-02-20
Target :input, not .input, in label macros

### LINKED ISSUE #397: Label macros can't be set to non-input elements
The `label-token` data attribute of a `Fieldmanager_Group` using `label_macro` is set, in part, to `input.fm-element`: https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/class-fieldmanager-group.php#L354. 

If I'm following correctly, this means that any non-input element, like a `<select>`, isn't found when `init_label_macros` attempts to update the label, so the macro isn't applied.

dlh01:
From the docs: "Selects all input, textarea, select and button elements." Fixes #397.


mboynes:
:+1:

PR #467, opened 2016-02-20
Address several bugs with save_to_post_parent

### LINKED ISSUE #283: PHP Notice or extra fields when Fieldmanager_Datasource_Post::$only_save_to_post_parent = true

mboynes:
Closes #307, closes #284, refs #18, resolves #283

### DIFF
--- a/php/class-fieldmanager-group.php
+++ b/php/class-fieldmanager-group.php
@@ -325,7 +325,7 @@ class Fieldmanager_Group extends Fieldmanager_Field {
 				elseif ( empty( $values[ $element->name ] ) ) unset( $values[ $element->name ] );
 			}
 
-			if ( ! empty( $element->datasource->only_save_to_taxonomy ) ) {
+			if ( ! empty( $element->datasource->only_save_to_taxonomy ) || ! empty( $element->datasource->only_save_to_post_parent ) ) {
 				unset( $values[ $element->name ] );
 				continue;
 			}
--- a/php/class-fieldmanager-group.php
+++ b/php/class-fieldmanager-group.php
@@ -148,6 +148,14 @@ class Fieldmanager_Group extends Fieldmanager_Field {
 				throw new FM_Developer_Exception( esc_html__( 'You cannot use `serialize_data => false` with `index => true`', 'fieldmanager' ) );
 			}
 
+			// A post can only have one parent, so if this saves to post_parent and
+			// it's repeatable, we're doing it wrong.
+			if ( $element->datasource && ! empty( $element->datasource->save_to_post_parent ) && $this->is_repeatable() ) {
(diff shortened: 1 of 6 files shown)


php/class-fieldmanager-group.php
```diff
@@ -148,6 +148,14 @@ public function __construct( $label = '', $options = array() ) {
 				throw new FM_Developer_Exception( esc_html__( 'You cannot use `serialize_data => false` with `index => true`', 'fieldmanager' ) );
 			}
 
+			// A post can only have one parent, so if this saves to post_parent and
+			// it's repeatable, we're doing it wrong.
+			if ( $element->datasource && ! empty( $element->datasource->save_to_post_parent ) && $this->is_repeatable() ) {
```
xmattus: Should this be `$element->is_repeatabe()` to check if the child element you're considering on this loop iteration is itself repeatable? You had checked `$this->is_repeatable()` back in class-fieldmanager-field.php

tests/php/test-fieldmanager-datasource-post.php
```diff
@@ -326,4 +325,111 @@ public function test_alter_child_invalid_reciprocal() {
...
+	 */
+	public function test_repeatable_post_parent_invalid() {
+		$fm = new Fieldmanager_Autocomplete( array(
+			'name'       => 'test_limitless_datasource',
+			'limit'      => 0,
+			'datasource' => new Fieldmanager_Datasource_Post( array(
+				'only_save_to_post_parent' => true,
+				'query_args' => array( 'post_type' => 'post' ),
+			) ),
+		) );
```
xmattus: Should there be assertions in this (and the next two tests)?

php/class-fieldmanager-group.php
mboynes: No, because the field itself is already checked at `Fieldmanager_Field::__construct()`, and the field doesn't have a parent yet (that happens after this loop).

tests/php/test-fieldmanager-datasource-post.php
xmattus: Cool

Issue #447, opened 2016-02-19
Update Handsontable


mboynes:
We're currently using 0.7.0-beta, current version is 0.23.0.

mslinnea:
This is used for Fieldmanager_Grid which will likely be deprecated.

PR #468, opened 2016-02-20
fieldmanager ui update

stevenslack:
- greater spacing between fields
- repeatable / sortable UI improvements
- media upload field provides more info about file uploaded
- tab UI updates
- new field CSS class which wraps just the label, field and description
- labels and descriptions style updates


mboynes:
* [ ] Test descriptions in horizontal tab groups
* [ ] Test descriptions in vertical tab groups
* [ ] Test descriptions in nested groups (2 and 3 levels deep)

css/fieldmanager-group-tabs.css
```diff
@@ -89,13 +116,14 @@
 
 .fm-wrapper .wp-tabs-panel > .fm-wrapper .fm-group-inner {
 	border-top: solid 1px #dfdfdf;
+	border-right: none;
+	border-bottom: none;
+	border-left: none;
+	-webkit-box-shadow: none;
```
ostowe: At some point it'd probably be nice to set up webpack for FM (including autoprefixer).

css/fieldmanager-group-tabs.css
```diff
@@ -3,9 +3,33 @@
 }
 
 .fm-tab-bar .fm-tab {
-	border: 1px solid transparent;
+	border: 1px solid #dfdfdf;
 	border-bottom: none;
+	padding: 8px 10px;
+	border-radius: 3px 3px 0 0;
+	background: #f1f1f1;
+	margin-right: 5px;
```
ostowe: On reviewing this feature branch with FM Demos I noticed both vertical and horizontal tabs do not work well at mobile widths. Horizontal tabs break strangely and overlap, vertical tabs leave very little space for the actual tab content. Does FM offer mobile support in general?

css/fieldmanager.css
```diff
@@ -30,6 +42,37 @@
 	margin: 0 0 0 5px;
 }
 
+.fm-label label {
+	font-size: 13px;
+	font-weight: 700;
+	line-height: 1.4em;
```
ostowe: In general I'd recommend unitless values for `line-height`, as they will work regardless of the unit you use for `font-size`.

css/fieldmanager.css
```diff
@@ -359,12 +614,13 @@ a.fm-delete:hover {
 	width: 25px;
 }
 
-.form-field input[type="checkbox"],
-.form-field input[type="radio"],
+.form-field input[type='checkbox'],
+.form-field input[type='radio'],
```
ostowe: Radio buttons are too close together on mobile.

css/fieldmanager.css
```diff
@@ -274,26 +453,26 @@ div.fm-group-label-wrapper .fmjs-remove {
...
 }
 
 .grid-toggle-wrapper.with-grid {
 	padding: 5px;
-	background: #fff;
 	border: solid 2px #ccc;
+	background: #fff;
 }
 
 .grid-activate {
```
ostowe: Seems like it's unrelated to these new styles, but the input styles for grid aren't working well. If you try to input data into a grid (at least in Chrome) you'll see what I mean`.handsontableInputHolder` needs to have a width set.

ostowe:
Styles are looking good over all, but had a few comments mostly related to mobile display.

css/fieldmanager-group-tabs.css
stevenslack: Yes, we'd also like to set up SASS which will allow for a better organization of CSS

css/fieldmanager.css
stevenslack: Thanks Owen, that has been flagged as an issue and will be addressed https://github.com/alleyinteractive/wordpress-fieldmanager/issues/447

css/fieldmanager.css
stevenslack: Addressed this issue in 805b9f6 and 1f9ab7e

mslinnea:
@stevenslack Overall, I really like these changes, but I did find a bug!  

It's probably not a common use case, but in one theme I worked on we have several places where there are double-nested Fieldmanager_Groups.  That's a Fieldmanager Group inside of a Fieldmanager Group inside of a Fieldmanager Group!  Actually the admin interface is pretty complicated, so there could be something else going on, but these changes are breaking the tabs UI of the top level group.  I'll send you a screenshot.

stevenslack:
@mslinnea addressed the issue you described. There were some edge cases where the sortable field handler was not positioned correctly in deeply nested group fields.

mslinnea:
@stevenslack Looks good to me!  Thanks for addressing those edge cases 😀  These UI improvements really do improve the look of repeatable fields.

css/fieldmanager-group-tabs.css
```diff
@@ -171,3 +188,29 @@
 	-webkit-box-shadow: none;
 	box-shadow: none;
 }
+
+@media screen and (min-width: 575px) {
+
+	.fm-tabbed-vertical .fm-tab-bar {
+		width: 175px;
+		float: left;
+		margin: 0 0 0 -12px;
```
montchr: I've checked this out locally and the `-12px` value causes vertical tabs to be offset incorrectly:

image 2018-05-03 at 11 58 29

Changing this to `margin: 0 0 0 1px;` fixes the issue for me. The `1px` is to account for the border of a field group:

image 2018-05-03 at 12 01 37

css/fieldmanager-group-tabs.css
```diff
@@ -89,13 +118,27 @@
...
+#fm_meta_box_vertical_tabbed_meta_fields .inside {
+	margin-top: 0;
+	padding: 0;
 }
+
+#fm_meta_box_vertical_tabbed_meta_fields .fm-group-description {
+	padding: 0 12px 12px;
+}
+
+#fm_meta_box_vertical_tabbed_meta_fields .wp-tabs-panel > .fm-wrapper .fm-group-inner {
```
montchr, line 138: I see the original selector was `.fm-tabbed-vertical` – any particular reason that would no longer work? I'd try to avoid ID selectors if possible, as I would think there could be more than one of these metaboxes per page?.

css/fieldmanager.css
```diff
@@ -30,6 +47,44 @@
...
+	line-height: 1.4;
+	display: inline-block;
+	margin: 0 0 4px;
+}
+
+.fmjs-removable-label label {
+	line-height: 1.7em;
+}
+
+/* Each FM Field */
```
montchr, line 62: Consider using a more prominent section divider comment if this refers to a large chunk of what follows. Perhaps [Idiomatic CSS comment style](https://github.com/necolas/idiomatic-css#comments) would be useful? Example from the linked page:

```css
/* ==========================================================================
   Section comment block
   ========================================================================== */

/* Sub-section comment block
   ========================================================================== */

/**
 * Short description using Doxygen-style comment format
 *
 * The first sentence of the long description starts here and continues on this
 * line for a while finally concluding here at the end of this paragraph.
 *
 * The long description is ideal for more detailed explanations and
 * documentation. It can include example HTML, URLs, or any other information
 * that is deemed necessary or useful.
 *
 * @tag This is a tag named 'tag'
 *
 * TODO: This is a todo statement that describes an atomic task to be completed
 *   at a later date. It wraps after 80 characters and following lines are
 *   indented by 2 spaces.
 */

/* Basic comment */
```

css/fieldmanager.css
```diff
@@ -30,6 +47,44 @@
...
+.fm-option input[type='checkbox'] {
+	margin-bottom: 7px;
+}
+
+/* Gallery */
+.fm-gallery .fm-item-description {
+	margin: 4px 0 10px;
+}
+
 /* Groups */
```
montchr, line 88: See previous note about more distinguishable section divider comments.

css/fieldmanager.css
```diff
@@ -95,94 +203,208 @@
...
 	margin: 0;
-	padding: 3px 5px 0 5px;
+	padding: 8px 12px 8px 12px;
 	border-bottom: none;
-	float: left;
+	background-color: transparent;
+	background-image: none;
 }
 
 /* Sortables and repeatables */
```
montchr, line 234: See previous comment about section dividers.

montchr (changes requested):
Looks good to me overall! I have one blocking question about the use of ID selectors in one place, and some suggestions about using more prominent section divider comments in CSS.

css/fieldmanager-group-tabs.css
stevenslack: yeah, this predates my memory. I agree.

css/fieldmanager.css
stevenslack: I agree although the basic comments here are just following the pattern in the rest of the stylesheets.

montchr (approved):
🌀

js/media/fieldmanager-media.js
```diff
@@ -78,15 +78,21 @@ $( document ).on( 'click', '.fm-media-button', function( event ) {
...
-			var preview = 'Uploaded file:<br />';
 			preview += wp.media.string.image( props );
+		} else if ( attachment.type == 'audio' ) {
+			preview += '<a href="#"><span class="dashicons dashicons-media-audio"></span></a>';
+		} else if ( attachment.type == 'video' ) {
+			preview += '<a href="#"><span class="dashicons dashicons-media-video"></span></a>';
 		} else {
-			var preview = 'Uploaded file:&nbsp;';
-			preview += wp.media.string.link( props );
+			preview += '<a href="#"><span class="dashicons dashicons-media-document"></span></a>';
```
dlh01, line 90: Is there a risk of these media icon classes becoming out of sync with the filtered icon class used in PHP?

php/class-fieldmanager-field.php
```diff
@@ -746,9 +757,14 @@ public function preload_alter_values( $values ) {
 	public function wrap_with_multi_tools( $html, $classes = array() ) {
 		$classes[] = 'fmjs-removable';
 		$out       = sprintf( '<div class="%s">', implode( ' ', $classes ) );
+		$handle = '';
 		if ( $this->sortable ) {
-			$out .= $this->get_sort_handle();
+			if ( ( $this->one_label_per_item || ! empty( $this->label ) ) && ! in_array( 'fmjs-removable-label', $classes ) && empty( $this->description ) ) {
```
dlh01, line 762:
```suggestion
			if ( ( $this->one_label_per_item || ! empty( $this->label ) ) && ! in_array( 'fmjs-removable-label', $classes, true ) && empty( $this->description ) ) {
```

php/class-fieldmanager-media.php
```diff
@@ -154,22 +154,69 @@ public function presave( $value, $current_value = array() ) {
...
+				$preview .= '<a href="#">';
+				$image_type = strpos( $attachment->post_mime_type, 'image/' );
+
+				if ( strpos( $attachment->post_mime_type, 'audio/' ) === 0 ) {
+					$dashicon_class = 'dashicons-media-audio';
+				} elseif ( strpos( $attachment->post_mime_type, 'video/' ) === 0 ) {
+					$dashicon_class = 'dashicons-media-video';
+				} else {
+					$dashicon_class = 'dashicons-media-document';
+				}
```
dlh01, line 172: `wp_attachment_is()` might be preferable here compared to checking the `post_mime_type` directly, as it will allow filters such as `wp_audio_extensions` to be used for sites allowing additional MIME types.

php/class-fieldmanager-media.php
```diff
@@ -154,22 +154,69 @@ public function presave( $value, $current_value = array() ) {
...
+
+				$dashicon_class = apply_filters( 'fieldmanager_media_preview_icon', $dashicon_class, $attachment->post_mime_type );
```
dlh01, line 174: This new filter should get filter docs.

Also noting in passing that I'm not sure there would be a way to target the icon for a specific field or a specific attachment with this filter. I'm also not sure what use cases there would be for it. All the same, it might be worth keeping in the back of your head how such abilities might be provided by either new field properties or additional context passed to the filter.

dlh01:
Couple potential edge cases I noticed while testing this branch on a client site:

Checkbox with a long label in the sidebar of the block editor breaks to a new line:

Screen Shot 2021-01-18 at 11 45 43 PM

Checkboxes within a tabbed group start to the left of the tab (not actually sure whether this is a defect):

Screen Shot 2021-01-18 at 11 45 33 PM

These cases aside, plus the few notes inline, the branch is looking really good to me. I'm excited to get it merged! Let me know what you think about the comments and how it'd be best to handle them, including in a follow-up PR.

js/media/fieldmanager-media.js
stevenslack: @dlh01 it is possible. I pushed up a change that will set the icon depending on the attachment icon type. See https://github.com/alleyinteractive/wordpress-fieldmanager/pull/468/files#diff-7f0fb55a63140bb4d44360770c1ec06ff18be39f6c9e359f8f7370bca480efcbR81-R83

Let me know what you think about his change.

php/class-fieldmanager-media.php
stevenslack: @dlh01 This filter was created so long ago that I can't recall the reasoning behind creating it. That said I am not sure this filter would provide that much value. Even if one was able to target for a specific field or attachment, would someone go through the effort in providing a different dashicon icon for an attachment? I am unsure and think that the filter should be removed unless there is a demand for such a feature.

dlh01: I like the approach of removing the filter 👍

stevenslack:
@dlh01 I addressed your feedback for the checkboxes wrapping to the next line in [3b6e936](https://github.com/alleyinteractive/wordpress-fieldmanager/pull/468/commits/9dd2cc2206a979537d69be2b82f6e01f43b6e936). For the checkboxes in the tabs, this isn't really a defect as much as it is just the default behavior.

js/media/fieldmanager-media.js
stevenslack: We are now relying on the media type from the attachment to output the correct icon.

dlh01:
@stevenslack Re. 3b6e936, where does the class `.fm-checkbox` come from? On the client site where that screenshot originated, the class isn't present, so the fix has no effect.

stevenslack:
@dlh01 The wrapper element, `fm-item` does not get updated with a class representing the type of field which it encapsulates. The commit [9dd2cc2](https://github.com/alleyinteractive/wordpress-fieldmanager/commit/9dd2cc2206a979537d69be2b82f6e01f43b6e936) is incorrect in this case. 

The issue you are referring to where a checkbox with a long label in the sidebar of the block editor is present on `master` as well and is not caused by this feature branch. This issue should be addressed in a separate UI update for FM.

dlh01:
Interestingly, I see that an instance of `.fm-checkbox` was added to the codebase in 2013: https://github.com/alleyinteractive/wordpress-fieldmanager/commit/cb4fe88f21f07b586d5a7e24d5a22f4d1ac4c846. Perhaps there used to be such a class-based class added.

As for the checkbox label, it looks as expected to me in `master`:

Screen Shot 2021-02-26 at 4 30 17 PM

Regardless, if you'd prefer to move the discussion to a separate issue, let me know.

stevenslack:
@dlh01 given the age on this PR I'd like to move the discussion of the checkbox label wrapping to another issue.

dlh01 (approved):
🌳 Let's go!

PR #476, opened 2016-02-23
Update the phpdoc headers for all classes

mboynes:


php/class-fieldmanager-richtextarea.php
```diff
@@ -1,9 +1,9 @@
 <?php
 
 /**
- * Use the WordPress Editor in Fieldmanager.
+ * TinyMCE visual editor which submits HTML.
```
nschoenholtz: It may be worth keeping a reference to the fact that we use the WP core editor implementation? Or if we'd rather keep it abstract it may make sense to just call it a WYSIWYG or something.

php/class-fieldmanager-autocomplete.php
```diff
@@ -1,8 +1,13 @@
 <?php
 
 /**
- * Post auto-complete field
- * @package Fieldmanager
+ * Text field that respond to user input with autocomplete suggestions
```
nschoenholtz: just a typo: responds

nschoenholtz:
A couple of minor comments but looks good to me!

mboynes:
Thanks @nschoenholtz!

mboynes:
Updated api.fieldmanager.org

PR #478, opened 2016-02-24
Add filter for fm_calculated_context() to support additional contexts and overrides

sc0ttkclark:
Includes hook docs! :)

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -367,24 +367,24 @@ function fm_calculate_context() {
 
 	if ( empty( $calculated_context ) ) {
 		$calculated_context = array( null, null );
-	}
 
-	/**
-	 * Filter the calculated context to allow for custom contexts and overrides to be used.
-	 *
-	 * @since 1.0.0
-	 *
-	 * @param array $calculated_context {
-	 *     Array of context information.
-	 *
-	 *     @type  string|null A Fieldmanager context of "post", "quickedit", "term",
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Let's get this into 1.0.0 before it's too late :)

danielbachhuber:
> Let's get this into 1.0.0 before it's too late

What are you trying to use it for? Can you share more details on your use case?

sc0ttkclark:
I was hoping to utilize this in additional contexts not yet supported, although it started out as me not fully understanding submenus covered settings pages. I thought this might be a useful hook in the future to help aide in integration with custom REST API endpoints (to tell FM what the context is, since it may not detect it).

bcampeau:
There's already a branch for REST API integration that adds a custom context function. That will supersede this.

Submenus pages are already covered too. I think we should likely close this.

sc0ttkclark:
Not including a filter like this would not promote integration testing in the future, just saying.

mboynes:
I like the idea in general, but it's too easy to inadvertently break things using this filter. I'm going to sleep on it and look at the code tomorrow; I think there's another place we could do this that would be safer.

sc0ttkclark:
What if we run it only if calculated context is empty?

dlh01:
It looks to me like this was added in #510.

mboynes:
Closing, as this was added in #510 as david noted.

PR #479, opened 2016-02-25
Added character counters

enrico-sorcinelli:
This PR adds character counter inside input text and textarea FM elements for all fields having `data-character-counter` attribute (see image below). For example:

```php
   'name' => 'foo',
   attributes' => [
      'data-character-counter' => true,
      'maxlength' => 100
   ]
) );
```
If the element has the `maxlength` attribute, then the counter shows the remaining characters instead of those typed.

The PR consists only of JS/CSS code and it exports the JavaScript `fm_add_character_counter` function that can be used not only within FM fields but also with all WP standard post's fields. For example the following code:

```javascript
fm_add_character_counter({ selector: '#title' });
```

add the counter to title input fields if applied in the edit post screen.  

PS: The default CSS values are for demonstration purposes and can be improved/modified.

fm_character_count

### DIFF
--- a/css/fieldmanager.css
+++ b/css/fieldmanager.css
@@ -369,3 +369,22 @@ a.fm-delete:hover {
 .form-field .fm-checkbox label {
 	display: inline;
 }
+
+/* character counter */
+.fm-character-counter {
+	position: relative;
+	display: inline;
+}
+
+.fm-character-counter span {
+	position: absolute;
+	right: 10px;
+	top: -1px;
+	opacity: 1;
+	color: #aaa;
+	font-size: 0.9rem;
+}
+
+.fm-character-counter-padding {
+	padding-right: 35px !important;
+}
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -267,4 +267,65 @@ $( document ).ready( function () {
 	$( document ).on( 'fm_activate_tab', init_sortable );
 } );
 
+/**
+ * Add character counter to input text / textarea elements
+ *
+ * @param object
+ *  - selector string jQuery selector
+ *  - type string It ca be 'lenght' (displays typed chars) or 'maxlength' (displays remaining chars)
+ */
+fm_add_character_counter = function() {
+	var args = jQuery.extend( {
+		debug: 0,
+		selector: null,
+		type: 'maxlength',
+		maxlength: 100,
+		force_container_block: false
+	}, arguments[0] );
+
+	jQuery( args.selector ).each( function(){
+
+		var el = jQuery( this );
+
+		// Check for fm
+		if ( el.closest( '.fm-item' ).hasClass( 'fmjs-proto' ) ) {
+			return;
+		}
+
+		// Add padding
+		el.addClass( 'fm-character-counter-padding' );
+
+		// Check type of counter
+		var type = typeof( el.attr( 'maxlength' ) ) === 'undefined' ? 'length' : args.type;
+
+		var div = jQuery( '<div />' );
+		var span = jQuery( '<span />' );
+
+		// Fix for textarea
+		if ( /^textarea$/i.test( el.prop( 'tagName' ) ) ) {
+			span.css( 'top', 'inherit' );
+			if ( args.force_container_block )
+				div.css( 'display', 'block' );
+		}
+
+		div.addClass( 'fm-character-counter' );
+		el.wrap(div);
+
+		span.text( type === 'length' ? el.val().length : el.attr( 'maxlength' ) - el.val().length );
+
+		el.before( span );
+
+		el.on( 'keydown', function() {
+			span.text( type === 'length' ? el.val().length : el.attr( 'maxlength' ) - el.val().length );
+		});
+		el.on( 'keyup', function() {
+			span.text( type === 'length' ? el.val().length : el.attr( 'maxlength' ) - el.val().length );
+		});
+	});
+}
+// Add character counter based on 'data-character-counter' attribute
+jQuery( document ).ready( function() {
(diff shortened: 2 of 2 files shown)


css/fieldmanager.css
```diff
@@ -369,3 +369,22 @@ a.fm-delete:hover {
...
+	position: absolute;
+	right: 10px;
+	top: -1px;
+	opacity: 1;
+	color: #aaa;
+	font-size: 0.9rem;
+}
+
+.fm-character-counter-padding {
+	padding-right: 35px !important;
```
danielbachhuber: Can we avoid `! important` here?

js/fieldmanager.js
```diff
@@ -267,4 +267,65 @@ $( document ).ready( function () {
...
+
+		// Check type of counter
+		var type = typeof( el.attr( 'maxlength' ) ) === 'undefined' ? 'length' : args.type;
+
+		var div = jQuery( '<div />' );
+		var span = jQuery( '<span />' );
+
+		// Fix for textarea
+		if ( /^textarea$/i.test( el.prop( 'tagName' ) ) ) {
+			span.css( 'top', 'inherit' );
```
danielbachhuber: Why can't this just be in CSS?

js/fieldmanager.js
```diff
@@ -267,4 +267,65 @@ $( document ).ready( function () {
...
+			if ( args.force_container_block )
```
danielbachhuber: Missing braces.

js/fieldmanager.js
```diff
@@ -267,4 +267,65 @@ $( document ).ready( function () {
...
+		}
+
+		div.addClass( 'fm-character-counter' );
+		el.wrap(div);
+
+		span.text( type === 'length' ? el.val().length : el.attr( 'maxlength' ) - el.val().length );
+
+		el.before( span );
+
+		el.on( 'keydown', function() {
```
danielbachhuber: These two `on` events need to be debounced, otherwise they'll slow perceived keyboard rate.

js/fieldmanager.js
```diff
@@ -267,4 +267,65 @@ $( document ).ready( function () {
...
+			span.text( type === 'length' ? el.val().length : el.attr( 'maxlength' ) - el.val().length );
+		});
+		el.on( 'keyup', function() {
+			span.text( type === 'length' ? el.val().length : el.attr( 'maxlength' ) - el.val().length );
+		});
+	});
+}
+// Add character counter based on 'data-character-counter' attribute
+jQuery( document ).ready( function() {
+	fm_add_character_counter( { selector: '.fm-element[data-character-counter]' } );
```
danielbachhuber: This also needs to be triggered on newly added elements, and potentially other custom actions.

danielbachhuber:
Thanks for taking this on, @enrico-sorcinelli !

For reference, here's the latest abstraction I've been using:

```
/**
 * Set up a counter on an element
 */
setupElementCounter: function( el ) {

	if ( ! el.length ) {
		return;
	}

	var wrap = $('<div />');
	wrap.addClass('frontline-counter-wrap');
	el.wrap( wrap );
	var span = $('<span />');
	span.addClass('frontline-counter');
	el.before( span );
	el.addClass('frontline-max-length-countdown');
	var setCounter = function() {
		var counter = el.val().length;
		if ( el.attr('maxlength' ) ) {
			counter = el.attr('maxlength') - counter;
		}
		span.text( counter );
		if ( el.attr('maxlength' ) && counter <= 10 ) {
			span.addClass('near-limit');
		} else {
			span.removeClass('near-limit');
		}
	}
	setCounter();
	el.on( 'keydown', this.debounce( setCounter ) );
},
```

Note the `near-limit` feature — for `maxlength`, it's nice to warn the user when they're getting close to the final limit, without discovering by hitting the hard stop.

css/fieldmanager.css
enrico-sorcinelli: As I wrote, even if the initial PR's intent was to add counter to FM fields, the `fm_add_character_counter()` function could add the counter also to all WP admin fields.
The `!important` clause allows to override the WP admin style without need to add the complete selectors for all IDs. For example for the title post we would have to add:

```css
#titlediv #title.fm-character-counter-padding {
	padding-right: 35px ;
}
```

js/fieldmanager.js
enrico-sorcinelli: Yes, of course.
The default CSS values were for demonstration purposes and can be improved/modified as the inline JS fixes. I just committed a fix to move them to the CSS file.
FYI the `display: block;` is for WP textarea fields.

js/fieldmanager.js
enrico-sorcinelli: Right

enrico-sorcinelli:
I was thinking to a similar *near-limit*  feature based on the percentage instead of hard code limit in order to avoid `near-limit` class always applyed for fields with `maxlenght` <= of that limit. For example:

```javascript
...
if ( el.attr('maxlength' ) && counter <= ( 20 / 100 ) * el.attr('maxlength' ) ) {
...
```
PS: For the maximum flexibility we could read the near-limit/percentage value from an additional field `data-attribute`

enrico-sorcinelli:
Hi @danielbachhuber, do you have any update on this?
I haven't found any reference about your abstraction (with the exception of the code you pasted here)

danielbachhuber:
> do you have any update on this?

No, I haven't worked on anything related recently.

> I haven't found any reference about your abstraction (with the exception of the code you pasted here)

Sorry, I should've clarified that the abstraction was meant just as a reference point, not something you needed to necessarily adopt.

PR #120, opened 2014-03-14
Allow Network Settings submenus and network-wide settings

dlh01:

PR #485, opened 2016-03-13
Add network submenu context

dlh01:
A replacement for #120 that extends `Fieldmanager_Context_Submenu` instead of sneaking network support into it. More details about the new approach are in the commit message in 37cc404.

I'm open to suggestions for adding multi-network support. I don't have any experience with multi-network installs, so I don't know what a good developer experience for them would look like. But it also seems like the context could move to `*_network_option()` in a backwards-compatible manner later.


dlh01:
I added this to the 1.1.0 milestone to match #120, but it definitely can be "future release" too.

PR #486, opened 2016-03-16
Add `fm-incrementable` class to hidden field in checkboxes

mboynes:


dlh01:
Could this be combined with #557?

mboynes:
Yeah, #557 supersedes this. Thanks!

Issue #490, opened 2016-04-18
Color Picker: Add a default color option


kjbenk:
I think it would make sense for the color picker to except a default color option so that users can understand and revert back to the default color at anytime.  This functionality is already handled within `wpColorPicker()` [here](https://github.com/WordPress/WordPress/blob/703d5bdc8deb17781e9c6d8f0dd7e2c6b6353885/wp-admin/js/color-picker.js#L149-L155).

mboynes:
I think this already works if you set the `data-default-color` attribute

kjbenk:
Correct you can set this attribute and the color picker will use a default color button instead of a clear button, but I wanted to create a way for the developers to specify this via the PHP constructor.  So that they can do something like this:

``` php
$fm = new Fieldmanager_Colorpicker( array( 'name' => 'default_colorpicker', 'default_color' => '#ff0000' ) );
```

I created a quick PR for this 😄

dlh01:
Added in #491.

PR #491, opened 2016-04-18
Adds a default color

kjbenk:
Allows developers to supply a new option `default_color` to specify whether or not the color picker will use the clear or default functionality.

From https://github.com/alleyinteractive/wordpress-fieldmanager/issues/490

### DIFF
--- a/php/class-fieldmanager-colorpicker.php
+++ b/php/class-fieldmanager-colorpicker.php
@@ -27,6 +27,13 @@ class Fieldmanager_Colorpicker extends Fieldmanager_Field {
 	public static $has_registered_statics = false;
 
 	/**
+	 * The default color for the color picker.
+	 *
+	 * @var string
+	 */
+	public $default_color = null;
+
+	/**
 	 * Build the colorpicker object and enqueue assets.
 	 *
 	 * @param string $label
@@ -43,6 +50,12 @@ class Fieldmanager_Colorpicker extends Fieldmanager_Field {
 
 		$this->sanitize = array( $this, 'sanitize_hex_color' );
 
+		if ( ! empty( $options['default_color'] ) ) {
+			$this->default_color = $options['default_color'];
+		} else if ( ! empty( $options['default_value'] ) ) {
+			$this->default_color = $options['default_value'];
+		}
+
 		parent::__construct( $label, $options );
 	}
 
@@ -57,7 +70,7 @@ class Fieldmanager_Colorpicker extends Fieldmanager_Field {
 			'<input class="fm-element fm-colorpicker-popup" name="%1$s" id="%2$s" data-default-color="%3$s" value="%4$s" %5$s />',
 			esc_attr( $this->get_form_name() ),
 			esc_attr( $this->get_element_id() ),
-			esc_attr( $this->default_value ),
+			esc_attr( $this->default_color ),
 			esc_attr( $value ),
 			$this->get_element_attributes()
 		);
--- a/php/class-fieldmanager-colorpicker.php
+++ b/php/class-fieldmanager-colorpicker.php
@@ -50,13 +50,13 @@ class Fieldmanager_Colorpicker extends Fieldmanager_Field {
 
 		$this->sanitize = array( $this, 'sanitize_hex_color' );
 
-		if ( ! empty( $options['default_color'] ) ) {
-			$this->default_color = $options['default_color'];
-		} else if ( ! empty( $options['default_value'] ) ) {
-			$this->default_color = $options['default_value'];
-		}
-
 		parent::__construct( $label, $options );
+
+		// If we have a default_value and default_color was not explicitly set
+		// to be empty, set default_color to default_value.
+		if ( ! isset( $this->default_color ) && ! empty( $this->default_value ) ) {
+			$this->default_color = $this->default_value;
+		}
 	}
 
 	/**
--- a/php/class-fieldmanager-colorpicker.php
+++ b/php/class-fieldmanager-colorpicker.php
@@ -20,6 +20,13 @@ class Fieldmanager_Colorpicker extends Fieldmanager_Field {
 	public $field_class = 'colorpicker';
 
 	/**
+	 * The default color for the color picker.
+	 *
+	 * @var string
+	 */
+	public $default_color = null;
+
+	/**
 	 * Static variable so we only load static assets once.
 	 *
 	 * @var boolean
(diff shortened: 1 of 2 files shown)


kjbenk:
@mboynes Please have a quick look at the regex I am using to make sure the default color is set.

srtfisher:
Curious about this: what is your rational for not using the `default_value` property?

kjbenk:
@srtfisher I am not sure I understand your question.  This PR is meant to allow developers to add a default color for the color picker itself.  This will change the `Clear` button to a `Default` button.

bcampeau:
I think this should use `default_value`

kjbenk:
@srtfisher I understand what you are referring to now 😄 

https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/class-fieldmanager-field.php#L233-L237

tests/php/test-fieldmanager-colorpicker-field.php
```diff
@@ -27,6 +27,15 @@ public function test_basic_render() {
 		$this->assertRegExp( '#<input class="[^"]*fm-colorpicker-popup[^>]+name="test_colorpicker"#', $html );
 	}
 
+	public function test_default_color() {
+		$fm = new Fieldmanager_Colorpicker( array( 'name' => 'test_colorpicker', 'default_value' => '#ff0000' ) );
+
+		ob_start();
+		$fm->add_meta_box( 'Test Colorpicker', 'post' )->render_meta_box( $this->post, array() );
+		$html = ob_get_clean();
+		$this->assertEquals( 1, preg_match( '[data-default-color="#ff0000"]', $html ) );
```
mboynes: You can use `$this->assertRegExp()` (see the test just above this one)

mboynes: while technically valid, square brackets are pretty uncommon for regex delimiters, especially since they're special characters in regex. The typically delimiter is `/`, except when that character is used in the regex, then you'll commonly find `#` used.

All that said, make this `'/data-default-color="#ff0000"/'`

mboynes:
The `$default_value` question is a good one and I had to think a lot about it. The question is: would you ever want the default in the color picker to be different from the initial value? There is one circumstance I can think of, which is where you want the value to be blank initially, but still be able to provide a default. This would especially be valid in repeating fields/groups, where setting a default value on a field will cause an otherwise empty field/group to save (e.g. when `extra_elements = 1`, which would cause a new field/group to be added with every save of the screen).

Here's my suggestion: Add a new variable `$default_color` as you did before. If `$default_value` is set and `$default_color` is not, set `$default_color` to `$default_value`. Be sure to _not_ set `$default_color` to anything, and use `isset()` to check it, that way someone could deliberately set the default color to an empty string if they wanted.

kjbenk:
@mboynes I implemented the method you have described above and added unit tests for each scenario.  Please let me know if that is what you envisioned.

php/class-fieldmanager-colorpicker.php
```diff
@@ -43,6 +50,12 @@ public function __construct( $label = '', $options = array() ) {
 
 		$this->sanitize = array( $this, 'sanitize_hex_color' );
 
+		if ( ! empty( $options['default_color'] ) ) {
```
mboynes: make this `isset()`

tests/php/test-fieldmanager-colorpicker-field.php
```diff
@@ -27,6 +27,32 @@ public function test_basic_render() {
...
+		ob_start();
+		$fm_2->add_meta_box( 'Test Colorpicker', 'post' )->render_meta_box( $this->post, array() );
+		$html = ob_get_clean();
+		$this->assertRegExp( '/data-default-color="#ff0000"/', $html );
+
+		$fm_3 = new Fieldmanager_Colorpicker( 'test-1', array( 'name' => 'test_colorpicker', 'default_color' => '', 'default_value' => '#ff0000' ) );
+		ob_start();
+		$fm_3->add_meta_box( 'Test Colorpicker', 'post' )->render_meta_box( $this->post, array() );
+		$html = ob_get_clean();
+		$this->assertRegExp( '/data-default-color/', $html );
```
mboynes: Should be:
```
$this->assertRegExp( '/data-default-color=""/', $html );
```

tests/php/test-fieldmanager-colorpicker-field.php
```diff
@@ -27,6 +27,32 @@ public function test_basic_render() {
...
+
+		$fm_4 = new Fieldmanager_Colorpicker( 'test-1', array( 'name' => 'test_colorpicker' ) );
+		ob_start();
+		$fm_4->add_meta_box( 'Test Colorpicker', 'post' )->render_meta_box( $this->post, array() );
+		$html = ob_get_clean();
+		$this->assertRegExp( '/data-default-color/', $html );
```
mboynes: Same as above

tests/php/test-fieldmanager-colorpicker-field.php
```diff
@@ -27,6 +27,32 @@ public function test_basic_render() {
...
+		ob_start();
+		$fm_1->add_meta_box( 'Test Colorpicker', 'post' )->render_meta_box( $this->post, array() );
+		$html = ob_get_clean();
+		$this->assertRegExp( '/data-default-color="#ff0000"/', $html );
+
+		$fm_2 = new Fieldmanager_Colorpicker( 'test-1', array( 'name' => 'test_colorpicker', 'default_value' => '#ff0000' ) );
+		ob_start();
+		$fm_2->add_meta_box( 'Test Colorpicker', 'post' )->render_meta_box( $this->post, array() );
+		$html = ob_get_clean();
+		$this->assertRegExp( '/data-default-color="#ff0000"/', $html );
```
mboynes: May as well also check value while you're at it...

```php
$this->assertRegExp( '/data-default-color="#ff0000" value="#ff0000"/', $html );
```

I guess there's no reason not to add this to the others, too

php/class-fieldmanager-colorpicker.php
kjbenk: My thinking was that since there is no real reason for a `''` default_color value, we should instead check for `! empty()` and not `isset()` to account for that.  In other words, when the user sets a default_color of `''` then we should change it to the default_value.  I hope that makes sense 😄

mboynes:
:+1:

PR #499, opened 2016-05-11
Fix argument order for optgroup key=>val pairs

mboynes:
This is a breaking change, but it was implemented wrong in the beginning. It's very rarely used, so I think it's safe to merge this.


scottnelle:
♻️

PR #508, opened 2016-07-20
Fieldmanager translations

enrico-sorcinelli:
This PR aims to load translation files for the plugin and adds italian translations.

I don't see an existing thread for it, as well as this feature doesn't appears on [1.1 road map](http://fieldmanager.org/2016/02/24/fieldmanager-1-1-kickoff/).

During translation I also fixed some very similar strings in order to reduce duplicated messages.

Moreover some enhancement to string messages should be done:

* Avoid to leave blank spaces at the end of messages:

For example in *php/class-fieldmanager-group.php* (line  ~ 145)

```php
esc_html( sprintf( __( 'Group child name conflict: %1$s / %2$s', 'fieldmanager' ), $name, $element->name ) );
```
could be better than:
```php
esc_html__( 'Group child name conflict: ', 'fieldmanager' ) . $name . ' / ' . $element->name );
```
* Normalize the syntax used in the messages:

Sometimes the double quotes (") were used to wrap code strings or values and sometimes not.
Sometime the backquotes (`) was used to wrap code and sometimes not.

Once we decided which syntax to use, I can work on this too.


dlh01:
@enrico-sorcinelli,

My plan to try to "break up the tasks" of preparing FM for translations hasn't succeeded, but I would appreciate your feedback on #571.

I tried to fix the issues with strings that you noticed, along with problems flagged by PHPCS, deviations from core, missing translator comments, etc. I hope the changes there would make it easier to write translations like yours.

mboynes:
This is currently blocked by #571, but I'd like to get it in in the 1.3.0 milestone.

PR #510, opened 2016-07-23
Filter the calculated context for plugins

mboynes:
Reviewing this PR would benefit from `?w=1` in the URL, as it's mostly whitespace.


netaustin:
🍂

PR #511, opened 2016-07-25
Fix bug in travis config, update WP

mboynes:
The way that Travis was testing phpunit, only the result of qunit indicated that the job passed or failed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: php
 matrix:
   include:
     - php: 5.3
-      env: WP_VERSION=4.3.3
+      env: WP_VERSION=4.4.4
     - php: 5.3
       env: WP_VERSION=latest
     - php: 5.6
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -39,7 +39,13 @@ script:
   - |
     if [[ ! -z "$WP_VERSION" ]] ; then
       phpunit
+    fi
+  - |
+    if [[ ! -z "$WP_VERSION" ]] ; then
```
kjbenk: Is there is specific reason why the phpunit needs WP_VERSION, when it is not needed in the starter theme? - https://github.com/alleyinteractive/wp-starter-theme/blob/master/.travis.yml#L126-L127

kjbenk:
Minor comment, otherwise :golf:

.travis.yml
mboynes: The travis config that we use internally is a little different. This is really aimed at future improvements to FM's travis config.

PR #513, opened 2016-07-27
.DS_Store is in gitignore

szepeviktor:
Deleted.


mboynes:
This was also deleted in #572. Thanks!

PR #517, opened 2016-08-08
Fix Bug with Saving Fields in PHP 7

mslinnea:
While using PHP Version 7.0.9 on the most recent Varying Vagrant Vagrants (VVV) environment, I found that field manager fields weren't saving.  I found that on line 49 of `class-fieldmanager-context-storable.php`, $data was an empty string instead of null.  Changing the check to empty fixed the issue I was experiencing.


mboynes:
We'll have to test this extensively before merging. Empty values (`false`, `0`, `""`, `"0"`, etc.) are valid, and we need to make sure those are all saved as expected.

mslinnea:
This turned out to be an issue unrelated to Fieldmanager.   The plugin `wp-cron-control` was setting DOING_CRON to true globally and that was causing `save_fields_for_cron()` to run instead of `save_fields_for_post()`

PR #525, opened 2016-08-25
Update valid_rules so other plugins/themes can extend validation. 

aaires:
In order to extend fields validation with some custom rules we would need to change the class variable `valid_rules`, so I purpose to have an `apply_filter` here in order to accomplish this.

### DIFF
--- a/php/util/class-fieldmanager-util-validation.php
+++ b/php/util/class-fieldmanager-util-validation.php
@@ -92,8 +92,10 @@ class Fieldmanager_Util_Validation {
 	 */
 	private function setup( $form_id, $context ) {
 		// Set class variables
-		$this->form_id = $form_id;
-		$this->context = $context;
+		$this->form_id     = $form_id;
+		$this->context     = $context;
+		//Allows other plugins or theme to extend available rules
+		$this->valid_rules = apply_filters('fieldmanager_valid_rules', $this->valid_rules);
 
 		// Add the appropriate action hook to finalize and output validation JS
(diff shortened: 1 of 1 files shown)


dlh01:
Hey @aaires,

Having a filter for the rules makes sense to me, but this code has been in FM for a while, so I'd be interested to hear from others who have been working on the plugin for longer than I have.

However, I also think any patch will need to follow WordPress coding standards about [spacing within parentheses](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage) and [filter documentation](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters).

dlh01:
Hey @aaires,

Sorry that this PR has been sitting for a while. 

After reading it again, I think it might be good to name the filter more specifically for `Fieldmanager_Util_Validation` -- off the top of my head, maybe `fm_util_validation_valid_rules` or something like that. 

The reason would be to try to avoid confusion with the `validation_rules` property of `Fieldmanager_Field`.

Also, can you add the multiline filter docs [as described in the WordPress documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters)?

dlh01:
Added to the 1.5 milestone in case anyone has a use for this and wants to carry it to completion, but if it doesn't get done for 1.5, it might be best to close for lack of interest.

PR #527, opened 2016-08-31
Cleanup validation JS and fix image validation rules #526

### LINKED ISSUE #526: Impossible to perform validation on Fieldmanager_Media
Hi there,

I am defining an image field with the Fieldmanager_Media with param validation_rules, in order to make it required.
`
/* Featured Image */
	$featured_image = new Fieldmanager_Media( false, array(
		'name'               => 'featured_image',
		'description'        => 'Feature Image(16:9) min size 800x450',
		'preview_size'       => 'medium',
		'label'              => 'Featured Image',
		'button_label'       => 'Add Image',
		'validation_rules' => array('required'=> true ),
	        'validation_messages' => array('required'=> 'Featured image is mandatory' ),
	) );`

This required validation rule is not running since in the code before  adding the Fieldmanager validation script `hidden` fields are ignored on purpose:

`	// Fields that should always be ignored
	$ignore[] = ".fm-autocomplete";
	$ignore[] = "input[type='button']";
	$ignore[] = ":hidden";`

Is it really necessary to include this hidden here? As far as I understand validation will only occur if validation_rules are specified. If this `hidden` is added to the `$ignore` array field of Fieldmanager_Media can't be required or have any other rules.

I am thinking of removing the ":hidden" from the `$ignore` array but would like some feedback in case you think this is compromising Fieldmanager in some way I am not currently foreseeing. 

Thank you in advance.

shadyvb:
- Fixes validation for image fields, which are hidden.
- Cleans up the validation JS, pass options to `wp_localize_script` instead of manually escaping and echo'ing every bit
- Pass options through a WP filter
- Pass options to JS handler in a way that allows Front-end extensions to work as well

Fixes #526

### DIFF
--- a/js/validation/fieldmanager-validation.js
+++ b/js/validation/fieldmanager-validation.js
@@ -1,22 +1,39 @@
+/*globals FM_VALIDATION_OPTIONS*/
 ( function( $ ) {
+	'use strict';
 
-fm_validation = {
-
-	invalidHandler: function( event, validator ) {
-		// Certain WordPress forms require additional cleanup to work smoothly with jQuery validation
-		var form = $( validator.currentForm ).attr( 'id' );
-		switch ( form ) {
-			case "post":
-				$( "#submitpost .spinner" ).hide();
-				$( "#submitpost #publish" ).removeClass( 'button-primary-disabled' );
-				$(window).off( 'beforeunload.edit-post' );
-				break;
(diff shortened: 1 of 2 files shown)


roborourke:
@mboynes any chance of some eyes on this one? Pretty simple short term fix for required media

PR #530, opened 2016-09-02
PHPCS for Datepicker and Checkboxes templates

dlh01:


templates/options-checkboxes.php
```diff
@@ -1,15 +1,24 @@
...
 	<input
 		class="fm-element"
 		type="checkbox"
 		value="<?php echo esc_attr( $data_row['value'] ); ?>"
 		name="<?php echo esc_attr( $this->get_form_name( '[]' ) ); ?>"
 		id="<?php echo esc_attr( $cb_id ); ?>"
-		<?php echo $this->get_element_attributes(); ?>
-		<?php echo $this->option_selected( $data_row['value'], $value, "checked" ); ?>
+		<?php echo $this->get_element_attributes(); // Escaped interally. xss ok. ?>
+		<?php echo esc_attr( $this->option_selected( $data_row['value'], $value, 'checked' ) ); ?>
```
nschoenholtz: i think this would output like, `checked=&quot;checked&quot;`, right?

nschoenholtz:
One concern (which if accurate, seems to reflect a missing unit test heh), otherwise yayyy ☑️

templates/options-checkboxes.php
dlh01: While I admit I didn't consider that possibility, as far as I can tell this returns just `checked`, not `checked="checked"`.

PR #531, opened 2016-09-03
PHPCS for assets and termmeta Utils

dlh01:


php/util/class-fieldmanager-util-term-meta.php
```diff
@@ -1,15 +1,19 @@
...
- * This is deprecated as of WordPress 4.4, which introduces Term Meta into core.
+ * @package Fieldmanager
+ */
+
+/**
+ * Use Fieldmanager to store meta data for taxonomy terms.
+ *
+ * This is deprecated as of WordPress 4.4, which introduces term MetA into core.
  * If you were using this feature prior to 4.4, it will continue to operate
  * until probably WordPress 4.8, however it is in your best interest to migrate
```
mslinnea: Curious, what's going to happen with WordPress 4.8 that would cause this to not work?

mslinnea:
🚅

php/util/class-fieldmanager-util-term-meta.php
dlh01: That is a better question for @mboynes, but my suspicion is that the comment means Fieldmanager probably will stop supporting `Fieldmanager_Util_Term_Meta` after 4.8.

mboynes: That's correct. That was just my estimation for when we'll officially deprecate it.

Issue #271, opened 2014-12-20
Adhere to WP Coding Standards


mboynes:
We are aiming to enforce the [WordPress coding standards](http://codex.wordpress.org/WordPress_Coding_Standards) across the entire plugin.

We'd love some help doing this spring cleaning. Below is a list of files that need to be verified or updated. If you want to chip in, comment on this issue saying that you're going to take one, and we'll update the issue to say you're on it. Then simply submit a PR with the clean file when you're done.
## PHP
- [x] `fieldmanager.php`
- [x] `php/class-fieldmanager-autocomplete.php`
- [x] `php/class-fieldmanager-checkbox.php`
- [x] `php/class-fieldmanager-checkboxes.php`
- [x] `php/class-fieldmanager-colorpicker.php`
- [x] `php/class-fieldmanager-datepicker.php`
- [x] `php/class-fieldmanager-draggablepost.php`
- [x] `php/class-fieldmanager-field.php`
- [x] `php/class-fieldmanager-grid.php`
- [x] `php/class-fieldmanager-group.php`
- [x] `php/class-fieldmanager-hidden.php`
- [x] `php/class-fieldmanager-link.php`
- [x] `php/class-fieldmanager-media.php`
- [x] `php/class-fieldmanager-options.php`
- [x] `php/class-fieldmanager-password.php`
- [x] `php/class-fieldmanager-radios.php`
- [x] `php/class-fieldmanager-richtextarea.php`
- [x] `php/class-fieldmanager-select.php`
- [x] `php/class-fieldmanager-textarea.php`
- [x] `php/class-fieldmanager-textfield.php`
- [x] `php/context/class-fieldmanager-context-page.php`
- [x] `php/context/class-fieldmanager-context-post.php`
- [x] `php/context/class-fieldmanager-context-quickedit.php`
- [x] `php/context/class-fieldmanager-context-storable.php`
- [x] `php/context/class-fieldmanager-context-submenu.php`
- [x] `php/context/class-fieldmanager-context-term.php`
- [x] `php/context/class-fieldmanager-context-user.php`
- [x] `php/context/class-fieldmanager-context.php`
- [x] `php/datasource/class-fieldmanager-datasource-post.php`
- [x] `php/datasource/class-fieldmanager-datasource-term.php`
- [x] `php/datasource/class-fieldmanager-datasource-user.php`
- [x] `php/datasource/class-fieldmanager-datasource.php`
[...]

PR #533, opened 2016-09-06
PHPCS for TextArea and Datepicker

dlh01:
See #271.

### DIFF
--- a/php/class-fieldmanager-datepicker.php
+++ b/php/class-fieldmanager-datepicker.php
@@ -1,64 +1,72 @@
 <?php
-
 /**
- * A Javascript date-picker which submits dates as unix timestamps.
+ * Class file for Fieldmanager_Datepicker.
  *
  * @package Fieldmanager_Field
+ */
+
+/**
+ * A JavaScript date-picker which submits dates as Unix timestamps.
  */
 class Fieldmanager_Datepicker extends Fieldmanager_Field {
 
 	/**
-	 * @var boolean
 	 * Collect time info or just date info? Defaults to just date info.
+	 *
+	 * @var bool
 	 */
-	public $use_time = False;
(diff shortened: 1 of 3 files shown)


php/class-fieldmanager-datepicker.php
```diff
@@ -93,23 +101,26 @@ public function form_element( $value ) {
...
+	 * @param mixed $current_value The current value for the field.
+	 * @return int Unix timestamp.
 	 */
 	public function presave( $value, $current_value = array() ) {
 		$time_to_parse = sanitize_text_field( $value['date'] );
 		if ( isset( $value['hour'] ) && is_numeric( $value['hour'] ) && $this->use_time ) {
 			$hour = intval( $value['hour'] );
 			$minute = ( isset( $value['minute'] ) && is_numeric( $value['minute'] ) ) ? intval( $value['minute'] ) : 0;
-			if ( $hour == 0 && $this->use_am_pm ) $hour = 12;
+			if ( 0 == $hour && $this->use_am_pm ) {
```
kjbenk: Should this use the strict comparison?

kjbenk:
Very minor comment, otherwise :golf:

php/class-fieldmanager-datepicker.php
dlh01: Possibly, but I don't want to risk changing the field's behavior in this PR, which is intended only to make the field standards-compliant.

kjbenk: Ok, but I thought using a strict comparison was making it standards compliant :)  Though it is only a warning so I guess it is fine.

mboynes: I don't see a problem with making that change now

PR #534, opened 2016-09-06
Fire action on field validation failure

dlh01:
As far as I know, it isn't currently possible to modify how individual fields respond to invalid input other than by extending `Fieldmanager_Field::_failed_validation()`. 

The goal of this PR is to add flexibility for contexts, or for other plugins or themes, in how they handle invalid input by firing an action in `_failed_validation()`.

Also hooks the new function `fm_failed_validation_to_wp_die()` to the new
'fm_failed_validation' action to mimic the previous behavior of calling `wp_die()` directly.


tests/php/test-fieldmanager.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Tests for functions in fieldmanager.php.
+ */
+
+class Test_Fieldmanager extends WP_UnitTestCase {
+	public function test_fm_failed_validation_to_wp_die() {
+		$rand_str = rand_str();
+
+		try {
+			fm_failed_validation_to_wp_die( $rand_str );
```
mboynes: If this doesn't `wp_die()`, this test will pass because the exception will not be thrown. I think the best bet is to set `@expectedException WPDieException` and after your assertion, `throw $e;`

fieldmanager.php
```diff
@@ -499,6 +499,23 @@ function fm_sanitize_textarea( $value ) {
...
+ *                              default message for validation errors.
+ */
+function fm_failed_validation_to_wp_die( $error_message = '' ) {
+	wp_die( esc_html(
+		$error_message
+		. "\n\n"
+		. __( "You may be able to use your browser's back button to resolve this error.", 'fieldmanager' )
+	) );
+}
+add_action( 'fm_failed_validation', 'fm_failed_validation_to_wp_die' );
```
mboynes: I'm not entirely sold on this, even though it successfully maintains backwards compatibility. The problems with dying are:

1. This field's data may be lost. One could argue that's OK, because it's invalid.
2. Fields hooked to save before this field (by Fieldmanager or not) **will** save.
3. Fields hooked to save after this field (by Fieldmanager or not) **will not** save and their data may be lost.

I'm at a bit of a loss on how best to handle this instead. I'm inclined to say that we should not save the invalid field by default, but make that filtered so one could elect to still save the invalid data (to prevent data loss). Then let all other fields save and do one of two things:

1. Hook into code later in execution (prior to the redirect) to die.
2. Let the request and redirect succeed, then on the next page render an admin notice with the error message. We could do this by setting a transient for the user with the error message, and filter the redirect to set a query arg to indicate that there was an error and that FM should look for it. Upon loading the next screen, if the query arg is present that there was an error, FM would load the transient and output the error message.

Of these ideas, I lean toward (2). However, I'd really love a better option.

bcampeau: I built a pretty cool admin notice framework for Apple News that we could look at using. It uses user meta instead of transients, but kind of amounts to the same. It doesn't require a query arg to trigger.

I like #2. I think it would be cool if it didn't save the data but displayed the invalid values in the admin notice. I'd also consider CSS to highlight the field(s) that had invalid values in red. They should display their values prior to the last save attempt though and not the invalid values.

roborourke: 2 is definitely better, it's better to collect errors, rely on sanitisation to avoid incorrect data being saved and finally to allow developers to decide what to do in the case of a validation error, most typical use case one I can think of would be keep a post as a draft if the requirements aren't met in that particular context.

A transient is how admin settings errors (added via `add_settings_error()` calls in the `register_setting()` sanitisation callback) are collected and displayed currently so I don't think it's such a bad option. You could also pass data to highlight the affected fields or add an error class.

Also I've implemented validation at the `wp_insert_post_data` hook before to avoid additional calls to `wp_update_post()` but you have to inspect the `$_POST` array rather than looking at post meta that's already been saved.

xmattus: Also voting 2. I can think of a number of weird circumstances where you really might need to work with the invalid data, so dying and discarding it seems to forestall that option unnecessarily.

bcampeau: Here's the admin notice class I referenced earlier, btw: https://github.com/alleyinteractive/apple-news/blob/master/admin/class-admin-apple-notice.php. I added some fairly specific output formatting recently, but just for reference.

shadyvb: Quick important note, pressing back in the browser to get back to editing does not actually work, there should probably be a link to the post edit page instead, to have consistent experience across browsers and different cases.

And (2) is what I actually opted for when I tried to solve this solution. I'm totally against dying, and even exceptions, IMHO exceptions should only be thrown for invalid code, not invalid user input, and that includes debug mode as well. Visual notices like admin messages are better suited to inform users of their erroneous input.

dlh01: @mboynes When you say "Then let all other fields save and do one of two things," would you mind clarifying what you mean by "other fields"?

shadyvb: I'm assuming he meant other fields of the form, because it doesn't save the rest of the fields if it dies in the middle parsing an invalid value of a certain field.

mboynes: Yes, as @shadyvb said -- what I meant by "other fields" is: any field hooked onto `save_post` after the invalid field, whether through Fieldmanager or not.

dlh01: All of the arguments against `wp_die()` seem convincing to me.

I think I get the relative advantages of option (2), but I want to make sure I understand it correctly. Would these points need to be addressed as part of implementing it?

- We would need some way of catching a validation error in field that is a child of a `Fieldmanager_Group` and blocking saving for the rest of the group (except for fields with `serialize_data => false`?).

- We would need to make sure we had a way to display the admin notice for fields that save over Ajax, like Quick Edit.

- Arguably, we would need a plan for displaying notices for contexts outside of the Dashboard. Although no such contexts currently exist in FM core (I don't thnk), the proposed user-forms (#157) and Customizer (#399) contexts come to mind.

sasimon:
I have managed to minimize this issue by leveraging Javascript and the included jQuery Validate. I pass in custom attributes and classes on the element (see [here](https://jqueryvalidation.org/jQuery.validator.addMethod/) on how to add a custom method, [here](https://jqueryvalidation.org/rules/) on how to set data-rule-* and data-msg-* attributes, and [here](https://jqueryvalidation.org/jQuery.validator.addClassRules/) to add a rule based on a CSS class). I have managed to translate many complex scenarios and needs by using jQuery to find related elements. In truth, I believe that documentation of the structure created by each field type is in order. Especially important is to indicate the classes created on each element, with {field_id} standing in for the field slug. Which I think is easier than addressing the issues posed by dh01 in his last post.

dlh01:
Closing for lack of traction. The approach outlined in https://github.com/alleyinteractive/wordpress-fieldmanager/pull/534#discussion_r77683276 would probably be best in a separate PR.

PR #536, opened 2016-09-14
Allow custom form templates

LC43:
This pull request allows usage of custom templates without having to change the plugin's core. It searches first in parent/child theme directory, for example in 

`[..]/themes/theme-name/plugins/fieldmanager/templates/`, 

and if it doesn't find the right template file, continue to search in plugin own directory.

Could be any path really. I chose this one arbitrarily .


LC43:
i'll review the code to see what has failed the CI.

mboynes:
There is already [an option to override templates on a per-field basis](http://api.fieldmanager.org/source-class-Fieldmanager_Field.html#221-225).

Issue #540, opened 2016-10-04
Fieldmanager_Autocomplete & Fieldmanager_Datasource_Term Bug for Taxonomy Array


mslinnea:
When I try to use a Fieldmanger_Datasource_Term that queries multiple taxonomies, it doesn't work.  Here's the code that doesn't work:

```
        $fm = new Fieldmanager_Group( array(
            'name'        => 'datasource_term',
            'children'    => array(
                'broken'         => new Fieldmanager_Autocomplete( array(
                    'label' => 'Test: Autocomplete with ajax',
                    'datasource' => new Fieldmanager_Datasource_Term( array(
                        'taxonomy' => array(  'category', 'post_tag' ),
                        'taxonomy_save_to_terms' => false
                    ) )
                ) ),
            )
        ) );
        $fm->add_meta_box( 'Test Datasource', 'demo-datasource' );
```

I think I may have found the bug.  PR coming soon.

mboynes:
There was no bug here, as noted in #541.

PR #541, opened 2016-10-04
Datasource_Term Bug: fix get_terms arguments

mslinnea:
`get_terms()` expects the first parameter to be an array with a `taxonomy` key.  This addresses issue #540 in part, as this fixed the example code in the post context, but I'm still experiencing an issue with the fm_widget context.


mboynes:
I believe this will break WP < 4.5, which I'd like to avoid if possible.

The change allowing the arguments array as the first param was put in with backwards compatibility support, so I'm surprised to see this behave differently. This may be something we want to trickle upstream to core if there's a bug.

Can you add unit tests for this? With those in place, we'll see how it goes.

Thanks!

mslinnea:
Nevermind, I'm having difficulty replicating the original issue.  But I did write a unit test, see #542, which passes and would seem to prove that I'm incorrect here.  Maybe I'll have a write a few more tests. 😄

mboynes:
It's been long enough at this point since 4.5 dropped, so we can move forward with this. Resolves #631.

PR #550, opened 2016-10-19
Update composer.json

s3rgiosan:
Defined the package type to `wordpress-plugin`. This means by default it will be installed into `wp-content/plugins` instead of the usual vendor directory.

Requiring `composer/installers` also means that anyone using this package can customize the install path as well (app/plugins for example).


mboynes:
Thanks! Sorry for the (very) delayed merge!

PR #556, opened 2016-11-22
Fire events when Colorpicker inputs change

dlh01:
This provides a degree of parity with sortable events, and it would allow context-specific callbacks where the color change needs to be reflected before data are saved. For an example, [see the use in this Customize context beta plugin](https://github.com/dlh01/fieldmanager-beta-customize/blob/1824d615a1c13d35a5fe8ab430fab83dfc1adf5e/js/fieldmanager-customize.js#L156-L177).


dlh01:
As with #560, there is a workaround via an existing event:

```js
$( document ).on( 'wpcolorpickercreate', '.fm-colorpicker-popup', function () {
	...
});
```

PR #560, opened 2016-11-26
Fire event when RichTextArea initializes

dlh01:
This event is in the spirit of #556, in that it allows other scripts to listen for relevant changes to an editor instance (for an example, [see the use in this Customize context plugin](https://github.com/dlh01/fieldmanager-beta-customize/blob/rte-events/js/fieldmanager-customize.js#L149-L162)). 

Unlike #556, it fires an event only on initialization, which is intended to reduce the number of added events and remain somewhat agnostic toward which editor is being initialized.


dlh01:
The `tinymce-editor-init` event already exists:

```js
$( document ).on( 'tinymce-editor-init', function ( event, editor ) {
    var editor_element = document.getElementById( editor.id );

    if ( editor_element && editor_element.classList.contains( 'fm-richtext' ) ) {
        ...
    }
});
```

which amounts to roughly the same thing as the proposed change in this PR, as far as I can tell, in which case there's less of a case for it.

Issue #555, opened 2016-11-20
Datepicker for international users


mbanusic:
Problem arises for users with interfaces with languages other than English. Datepicker uses `strtotime` which only translates English date format (see [docs](http://php.net/manual/en/function.strtotime.php)). Now since translated version of jQuery return translated months it causes problems, e.g. "20 stu. 2016" in Croatian is not parsable by `strtotime`.

For international users fix is to use some other date format, e.g. 20-11-2106, but since the plugin should work out of the box, we should either change the default date format, or implement some mechanism to handle international usage.

dlh01:
[The Datepicker's altField and altFormat options](http://api.jqueryui.com/datepicker/#option-altField) look like they're designed for a situation like this one.

dlh01:
@mbanusic Could you please test whether the branch in #562 addresses this issue?

mbanusic:
@dlh01 I tested #562 on my local installation and now it saves properly in the timestamp format while using the proper locale in datepicker. Thank you for the fix.

dlh01:
Thanks for testing. I think we should leave this issue open until a fix for it is released, though.

PR #562, opened 2016-12-18
Submit Datepicker value in language-agnostic format

dlh01:
See #555. Also adds localization to some parts of the default Datepicker template and bumps the QUnit version.

### DIFF
--- a/templates/datepicker.php
+++ b/templates/datepicker.php
@@ -31,8 +31,8 @@
 		<input class="fm-element fm-datepicker-time" type="text" value="<?php echo esc_attr( $this->get_minute( $value ) ); ?>" name="<?php echo esc_attr( $this->get_form_name( '[minute]' ) ); ?>" />
 		<?php if ( $this->use_am_pm ) : ?>
 			<select class="fm-element" name="<?php echo esc_attr( $this->get_form_name( '[ampm]' ) ); ?>">
-				<option value="am"<?php selected( $this->get_am_pm( $value ), 'am' ) ?>>A.M.</option>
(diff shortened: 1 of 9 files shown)


templates/datepicker.php
```diff
@@ -26,8 +31,8 @@ class="fm-element fm-datepicker-popup"
 		<input class="fm-element fm-datepicker-time" type="text" value="<?php echo esc_attr( $this->get_minute( $value ) ); ?>" name="<?php echo esc_attr( $this->get_form_name( '[minute]' ) ); ?>" />
 		<?php if ( $this->use_am_pm ) : ?>
 			<select class="fm-element" name="<?php echo esc_attr( $this->get_form_name( '[ampm]' ) ); ?>">
-				<option value="am"<?php selected( $this->get_am_pm( $value ), 'am' ) ?>>A.M.</option>
-				<option value="pm"<?php selected( $this->get_am_pm( $value ), 'pm' ) ?>>P.M.</option>
+				<option value="am"<?php selected( $this->get_am_pm( $value ), 'am' ) ?>><?php echo esc_html( $wp_locale->get_meridiem( 'AM' ) ) ?></option>
+				<option value="pm"<?php selected( $this->get_am_pm( $value ), 'pm' ) ?>><?php echo esc_html( $wp_locale->get_meridiem( 'PM' ) ) ?></option>
```
dlh01: Upon reflection, getting this translation from `$wp_locale` might be overkill because `esc_html_e( 'AM' )` (or 'PM') would have the same effect. But, it would also rely on that string remaining translated in core, and it would cause the plugin to draw from multiple text domains.

PR #565, opened 2016-12-20
Use `sanitize_textarea_field()` with TextAreas where available

### LINKED ISSUE #551: Consider replacing `fm_sanitize_textarea()` with `sanitize_textarea_field()` (where available)
WordPress 4.7 is on track to add a sanitizing function for textareas -- see https://core.trac.wordpress.org/changeset/38944.

It doesn't seem like `fm_sanitize_textarea()` could be removed entirely, but, if `sanitize_textarea_field()` makes it into 4.7, then FM could potentially defer to core when the function exists.

dlh01:
Closes #551.

### DIFF
--- a/php/class-fieldmanager-textarea.php
+++ b/php/class-fieldmanager-textarea.php
@@ -29,8 +29,11 @@ class Fieldmanager_TextArea extends Fieldmanager_Field {
 			'rows' => '10',
 		);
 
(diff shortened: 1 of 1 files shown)


mboynes:
@dlh01 to update to only use the WP4.7+ code.

Issue #523, opened 2016-08-24
The `fm_presave_alter_values` filter does not fire for Options fields


kjbenk:
Currently any field that is a subclass of the [Options Field](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/class-fieldmanager-options.php) does not have access to the `fm_presave_alter_values` filter.

dlh01:
Fixed in #568.

PR #568, opened 2016-12-22
Apply 'fm_presave_alter_values' to `Fieldmanager_Options`

dlh01:
First pass at #523 and a partial refresh of #394. However, is it a backwards-incompatibility to start applying this filter now?


mboynes:
While it technically is backwards-incompatible, that is a small price to pay. Proceed! 👍

dlh01:
As was noted in #394, the issue this PR addresses also applies to autocomplete fields. I've expanded the scope of this PR to include autocompletes, although those changes could be moved into a separate PR if you'd like.

mboynes:
I'm in favor of autocomplete being a part of this PR, thanks!

PR #569, opened 2016-12-23
A fix to handle showing CHozen items if they are loaded in a hidden block

pbearne:
There is a know problem https://github.com/harvesthq/chosen/issues/92 where Chozen doesn't load in hidden elements. 
This fix just rechecks for elements that are marked to have chozen added but don't have it loaded and triggers it to be loaded if found when the trigger_display_if is run

### DIFF
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -266,7 +266,7 @@ $( document ).ready( function () {
 				}
 			}
 		} );
-		// check for chozen not loaded do to being hidden and load t found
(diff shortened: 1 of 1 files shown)


pbearne:
this failing in travis with 'undefined' is not a function (evaluating '$(".chzn-select").not( ".chzn-done" ).chosen()') not sure why

pbearne:
this might not be a full fix I am finding that elements in a repeater that is hidden on load are set to width 0

pbearne:
added a reload for chosen on add_another element

PR #573, opened 2017-01-04
Remove reference to undefined $user_id in Fieldmanager_Context_Page::save_page_form()

montchr:
When calling `Fieldmanager_Context_Page::save_page_form()`, a notice is
thrown for the undefined `$user_id` variable. There's no need for
accessing a user ID here.


dlh01:
Hey @montchr,

I don't have a lot of experience with `Fieldmanager_Context_Page`, so I think you could say better than I could whether removing that assignment would affect backwards-compatibility.

If I'm following correctly, while it's true that there is no `$user_id` variable, the assignment means in practice that `$this->fm->data_id` will always be set to `null`, and so the condition later:

```
if ( empty( $this->fm->data_id ) ) $this->fm->data_id = $this->uniqid;
```

will always be true.

If we remove the assignment, is it possible that `$this->fm->data_id` would not be `empty()` and so might not be set to `$this->uniqid`? If so, are there any consequences?

dlh01:
With `Fieldmanager_Context_Page` deprecated since #653, we're going to close this one rather than try to sort out the potential backcompat consequences for any uses still out there.

PR #574, opened 2017-01-04
Define Composer package type

montchr:
This allows users to easily manage this plugin as a Composer dependency.

https://roots.io/wordpress-plugins-with-composer/


dlh01:
(See also #550.)

montchr:
My mistake, missed that one.

PR #575, opened 2017-01-04
Allow filtering of Post Datasource result titles

montchr:
It would be helpful to be able to filter the titles of each search result in a post datasource, for example to add a post type or post ID.

An example:


dlh01:
I wonder whether it would be possible to use the `fm_datasource_term_get_items` filter as a model for the name and passed arguments of this filter (for example, including `$fragment`). If so, would you still be able to use the filter as you did in your example?

montchr:
> I wonder whether it would be possible to use the `fm_datasource_term_get_items` filter as a model for the name and passed arguments of this filter (for example, including `$fragment`). If so, would you still be able to use the filter as you did in your example?

Yes totally. I tried out the approach in #696 and it works great. Closing this pull request in favor of that more robust approach.

PR #580, opened 2017-02-07
minimum_count: Exactly what it says on the tin

sasimon:
This commit will cause the remove handles on repeated groups to not work
if the count of the repeated group is less than or equal to the
minimum_count. This brings the minimum_count feature closer to working
as I believe it was intended, and less of a clone of features already
offered by starting_count and extra_elements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ todo.txt
 composer.lock
 /vendor/
 /node_modules
+*.js~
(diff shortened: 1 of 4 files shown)


sasimon:
I do not wish this pull request to be merged. I am working on a solution that achieves this with faster performance from the unit tests.

Issue #404, opened 2015-11-19
Surface field data in the REST API


danielbachhuber:
It would be neat if I could add a `show_in_rest=>true` argument to any field (or group) I've defined, and have it automatically exposed in the REST API.

mboynes:
Great idea. I've got another idea for FM and the REST API, so I'm going to rename this to disambiguate

haritpanchal:
Guys, do you have any updates regarding this issue? We are using the FM and it is the blocker for us. Waiting for your response! @danielbachhuber @mboynes

mogmarsh:
@haritpanchal what's your use case? Are you wanting fields in the REST API for admin use or for front end use? Do you need all the field data (field type, options, etc) or just the meta values?
I have started a [project](https://github.com/alleyinteractive/fm-gutenberg/) which converts FM fields into native Gutenberg controls, and as part of this makes all the field data and values available in the REST API for admin use, but not for front end use. You might be able to extract some of what you need from it.

mboynes:
@haritpanchal see [this comment](https://github.com/alleyinteractive/wordpress-fieldmanager/issues/507#issuecomment-993994370) for a complete explanation. At this point, it's safe to conclude that we won't merge #583 to provide this functionality, for the reasons I detailed in that comment. That shouldn't be a blocker for you, however, as adding REST support to any given post or term meta, or site option, is a trivial task. And doing so from a theme or custom plugin, as I noted in that comment, will be much more stable and future-looking than depending on Fieldmanager to register the data with the REST API.

I'm going to close this issue and that PR to finally remove the ambiguity.

haritpanchal:
@mogmarsh, We want to use the fields on the admin side, and we mostly need meta values.

Issue #507, opened 2016-07-12
Consider support for register_meta()


johnbillion:
[WordPress 4.6 introduces a fundamental change to the way meta fields are registered.](https://make.wordpress.org/core/2016/07/08/enhancing-register_meta-in-4-6/) The primary change is the addition of meta metadata such as its type, description, value structure, and whether to expose it in the REST API.

We should consider whether registered Fieldmanager meta fields should also call `register_meta()` in order to expose meta fields to this new meta API.

bcampeau:
This is actually in progress. Contributions welcome!

https://github.com/alleyinteractive/wordpress-fieldmanager/tree/rest_api

spacedmonkey:
@joemcgill  Any chance of adding register_meta support to fields? Adding show_in_rest support would be great!

mboynes:
For posterity, related issues and PRs: #404 #407 #583

PR #583, opened 2017-02-13
REST API: Expose fields to rest api

kjbenk:
This PR gives the ability for developers to expose their FM fields to the REST API.  Most of the work comes from Brad's branch [rest_api](https://github.com/alleyinteractive/wordpress-fieldmanager/tree/rest_api).

By default, no fields are exposed to the REST API.  In order to expose a single field to the REST API you just need to set the fields attribute of `show_in_rest` to `true` and add a new action for `fm_rest_api_{context}_{type}`

Example:
```php
/**
 * Add REST post field.
 */
function add_custom_field() {
	$fm = new Fieldmanager_TextField( [
		'name' => 'rest_field',
		'show_in_rest' => true,
	] );
	$fm->add_meta_box( __( 'Test Fields' ), array( 'post' ), 'side' );
}
add_action( 'fm_post_post', 'add_custom_field' );
add_action( 'fm_rest_api_post_post', 'add_custom_field' );
```


mboynes:
Resolves #404

kjbenk:
I am thinking through a better way to handle this, so I will put this PR on hold for now.

kjbenk:
I moved all `get` and `update` logic to the root context class since it provides better functions for gathering data like `load()` as opposed to a child class, like the post context, which will use `get_data()`.

dlh01:
Oh, and, of course, if I'm misunderstanding anything or misconstruing the changes in the branch, please tell me!

php/context/class-fieldmanager-context-submenu.php
```diff
@@ -198,6 +201,56 @@ public function url() {
...
 	}
 
+	/**
+	 * Register a new REST API endpoint to expose the field's data.
+	 *
+	 * @param string $type The fm field type.
+	 */
+	public function register_rest_field( $type = 'submenu' ) {
+		if (
+			function_exists( 'register_rest_route' )
```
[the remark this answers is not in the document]
kjbenk: If FM forces WP 4.7+ then this check can be removed.

php/context/class-fieldmanager-context-submenu.php
```diff
@@ -198,6 +201,56 @@ public function url() {
 		}
 	}
 
+	/**
+	 * Register a new REST API endpoint to expose the field's data.
+	 *
+	 * @param string $type The fm field type.
+	 */
+	public function register_rest_field( $type = 'submenu' ) {
```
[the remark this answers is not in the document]
kjbenk: This is needed since it is overriding the `register_rest_field` function within `Fieldmanager_Context`, which takes a single var.

fieldmanager.php
```diff
@@ -435,6 +440,93 @@ function fm_trigger_context_action() {
...
+ *
+ * This is separate from fm_trigger_context_action since rest_pre_dispatch fires on parse_request
+ * and is too late to be part of the existing context hooks.
+ *
+ * @param mixed           $result  Response to replace the requested version with. Can be anything
+ *                                 a normal endpoint can return, or null to not hijack the request.
+ * @param WP_REST_Server  $server  Server instance.
+ * @param WP_REST_Request $request Request used to generate the response.
+ */
+function fm_trigger_rest_context_action( $result, $server, $request ) {
```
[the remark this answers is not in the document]
kjbenk: Done :)

kjbenk:
@dlh01 

> Not honoring 'serialize_data' => false seems potentially problematic. On the other hand, given that the data isn't being represented as meta on the resource, maybe it's no big deal.

`serialize_data` is being handled for the `post`, `term` and `user` contexts since we are using the [load](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/context/class-fieldmanager-context-storable.php#L119-L130) function.

For the submenu context, `serialize_data` I believe is not used which is why I use `get_data` there.

> Adding top-level fields to objects returned by the API is sort of a break with Fieldmanager tradition. As I read it, it would be as if FM allowed developers to add properties to WP_Post, WP_Term, etc. objects. Not necessarily bad, just an observation.

I was thinking about that and I understand the concerns.  Adding all of the data to a `fm-meta` field might be worth looking into, but I would love to hear any solutions you might have.  Unless we are ok with outputting the data within the root node, either way I think this is a great discussion to have.

> The existing context classes are being used as a proxy for determining whether a a REST field should be added to an object type. However, historically, I think, FM contexts have been intended to correspond to locations where fields display, not 1:1 with object types. For example, I could have a field that appears only in the Quick Edit context, and the field data would be saved to post meta, but I don't think it would be represented in the REST API by this branch. Again, might not be a big deal, just something to consider.

Great point.  I am not totally sure how to best handle this or if we even need as you pointed out, so I might need to think this over and see how it can be done better.

> Similarly, the name of the submenu-settings endpoint strikes me as a little confusing. A "submenu" is a concept specific to wp-admin, but the underlying option (which the endpoint retrieves) isn't tied to wp-admin.

I totally agree. I was not 100% sure what to name the endpoint so I just used the name of the context as a starting point. Please let me know if any names you were thinking of using. Here are some off the top of my head:

- `/fm/v1/settings/{setting-name}`
- `/fm/v1/options/{setting-name}`

php/class-fieldmanager-field.php
```diff
@@ -1399,4 +1420,54 @@ public function escape( $field, $default = 'esc_html' ) {
...
+	/**
+	 * Determine if this field can be shown in the REST API.
+	 *
+	 * @since 1.3.0
+	 *
+	 * @return bool True or false if the field can be shown in the REST API.
+	 */
+	public function can_show_in_rest() {
+		/**
+		 * Determine whether this field should be shown within the REST API.
```
dlh01: Should use "Filters..." wording.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+	 */
+	public function rest_get_callback( $object, $field_name, $request, $object_type ) {
+		// Set the current ID.
+		$this->fm->data_id = $object['id'];
+
+		// Get the data.
+		$data = $this->load();
+
+		/**
+		 * Filters a single field's data to the REST API.
```
dlh01: Missing a word ("data _passed_ to"?).

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+
+		/**
+		 * Filters all data passed to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
+		 * @param  string          $field_name  The REST API field name.
+		 * @param  WP_REST_Request $request     The full request object from the REST API.
+		 * @param  string          $object_type The REST API object type
```
dlh01: Missing documentation for the field object.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+		/**
+		 * Filters a single field's data to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
+		 * @param  string          $field_name  The REST API field name.
+		 * @param  WP_REST_Request $request     The full request object from the REST API.
+		 * @param  string          $object_type The REST API object type
+		 * @param  object          $fm          The base fm object.
```
dlh01: "FM object"

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+		 * Filters all data passed to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
+		 * @param  string          $field_name  The REST API field name.
+		 * @param  WP_REST_Request $request     The full request object from the REST API.
+		 * @param  string          $object_type The REST API object type
+		 */
+		$data = apply_filters( 'fm_rest_update', $data, $object, $field_name, $request, $object_type, $this );
```
dlh01: Missing documentation for the field object.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+
+		// Get the data.
+		$data = $this->load();
+
+		/**
+		 * Filters a single field's data to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
```
dlh01: Extra space after the `@param`s in this filter.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+		 * @param  object          $fm          The base fm object.
+		 */
+		$data = apply_filters( 'fm_rest_get_' . $field_name, $data, $object, $field_name, $request, $object_type, $this );
+
+		/**
+		 * Filters all data passed to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
```
dlh01: Extra space after the `@param`s in this filter.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+	 * @param WP_REST_Request $request     The full request object from the REST API.
+	 * @param string          $object_type The REST API object type.
+	 */
+	public function rest_update_callback( $data, $object, $field_name, $request, $object_type ) {
+		/**
+		 * Filters a single field's data to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
```
dlh01: Extra space after the `@param`s in this filter.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+		 * @param  string          $object_type The REST API object type
+		 */
+		$data = apply_filters( 'fm_rest_update_' . $field_name, $data, $object, $field_name, $request, $object_type, $this );
+
+		/**
+		 * Filters all data passed to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
```
dlh01: Extra space after the `@param`s in this filter.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+	public function rest_update_callback( $data, $object, $field_name, $request, $object_type ) {
+		/**
+		 * Filters a single field's data to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param  array           $object      The REST API object.
+		 * @param  string          $field_name  The REST API field name.
+		 * @param  WP_REST_Request $request     The full request object from the REST API.
+		 * @param  string          $object_type The REST API object type
```
dlh01: Missing documentation for the field object.

php/context/class-fieldmanager-context-storable.php
```diff
@@ -182,6 +182,104 @@ public static function sanitize_scalar_value( $value ) {
...
+	 *
+	 * @param mixed           $data        The value to be updated for the field from the request.
+	 * @param object          $object      The REST API object.
+	 * @param string          $field_name  The REST API field name.
+	 * @param WP_REST_Request $request     The full request object from the REST API.
+	 * @param string          $object_type The REST API object type.
+	 */
+	public function rest_update_callback( $data, $object, $field_name, $request, $object_type ) {
+		/**
+		 * Filters a single field's data to the REST API.
```
dlh01: Missing a word ("data [...] to").

php/context/class-fieldmanager-context-submenu.php
```diff
@@ -198,6 +201,63 @@ public function url() {
...
+	 * @since 1.3.0
+	 *
+	 * @return mixed The fm field data.
+	 */
+	public function rest_callback() {
+		// Get the data.
+		$data = $this->get_data( 0, $this->fm->name );
+
+		/**
+		 * Filters a single field's data to the REST API.
```
dlh01: Missing a word ("data [...] to").

php/context/class-fieldmanager-context-submenu.php
```diff
@@ -198,6 +201,63 @@ public function url() {
...
+		/**
+		 * Filters all data passed to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param array  $data The field data.
+		 * @param array  $data The field name.
+		 * @param string $this The current context object.
+		 */
+		$data = apply_filters( 'fm_rest_get_submenu', $data, $this->fm->name, $this );
```
dlh01: Both the `fm_rest_get` and `fm_rest_get_submenu` filters are documented with "Filters all data passed to the REST API," but they appear to slightly more specific than "all" in practice.

php/context/class-fieldmanager-context-submenu.php
```diff
@@ -198,6 +201,63 @@ public function url() {
...
+		 */
+		$data = apply_filters( 'fm_rest_get_submenu_' . $this->fm->name, $data, $this );
+
+		/**
+		 * Filters all data passed to the REST API.
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param array  $data The field data.
+		 * @param array  $data The field name.
```
dlh01: `@param string  $name`

fieldmanager.php
```diff
@@ -435,6 +440,93 @@ function fm_trigger_context_action() {
...
+	// Store matches to use with the context action.
+	$matches = array();
+
+	// Use regexes to find the right context.
+	if (
+		! empty( $post_type_endpoints )
+		&& preg_match( '/\/wp\/v2\/(' . implode( '|', array_keys( $post_type_endpoints ) ) . ')(\/?)(.*?)/', $route, $matches )
+	) {
+		// Post Types.
+		do_action( "fm_post_{$post_type_endpoints[ $matches[1] ]}" );
```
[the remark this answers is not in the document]
dlh01: Bump on this question.

php/class-fieldmanager-options.php
```diff
@@ -347,4 +347,27 @@ protected function add_meta_boxes_to_remove( &$meta_boxes_to_remove ) {
...
+	 * Creates the JSON Schema for all options fields.
+	 *
+	 * @see http://json-schema.org/draft-04/schema#
+	 */
+	protected function create_schema() {
+		// Add the required schema properties.
+		parent::create_schema();
+
+		// Build data from the datasource, which is needed by the schema.
+		if ( ! $this->has_built_data && ! empty( $this->datasource ) ) {
```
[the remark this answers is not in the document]
dlh01: Bump on this question.

fieldmanager.php
```diff
@@ -435,6 +443,100 @@ function fm_trigger_context_action() {
...
+	// Store matches to use with the context action.
+	$matches = array();
+
+	// Use regexes to find the right context.
+	if (
+		! empty( $post_type_endpoints )
+		&& preg_match( '#/wp/v2/(' . implode( '|', array_keys( $post_type_endpoints ) ) . ')(/?)(.*?)/#i', $route, $matches )
+	) {
+		// Post Types.
+		do_action( "fm_post_{$post_type_endpoints[ $matches[1] ]}" );
```
dlh01: Duplicate hooks in this file should be documented per the [core docs standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-1-duplicate-hooks).

php/context/class-fieldmanager-context.php
```diff
@@ -100,4 +100,33 @@ protected function render_field( $args = array() ) {
...
+	 */
+	public function register_rest_field( $object_type ) {
+		// Ensure the REST API is active and the field wants to be shown in REST.
+		if ( $this->fm->can_show_in_rest() ) {
+			register_rest_field(
+				$object_type,
+				$this->fm->name,
+				array(
+					'get_callback'    => array( $this, 'rest_get_callback' ),
+					'update_callback' => array( $this, 'rest_update_callback' ),
```
dlh01: These callbacks are no longer part of this class.

php/class-fieldmanager-field.php
```diff
@@ -1399,4 +1420,54 @@ public function escape( $field, $default = 'esc_html' ) {
...
+		 *
+		 * @since 1.3.0
+		 *
+		 * @param bool               $this->show_in_rest The current show in rest boolean.
```
dlh01: Omit `$this` in the param docs -- just `$show_in_rest`.

php/class-fieldmanager-field.php
```diff
@@ -1399,4 +1420,54 @@ public function escape( $field, $default = 'esc_html' ) {
...
+		 * @param Fieldmanager_Field                     The current field object.
+		 */
+		return apply_filters( 'fm_can_show_in_rest', $this->show_in_rest, $this );
```
dlh01: The more I think about it, the more I wonder whether the this filter is an invitation to accidentally expose sensitive data. All a developer would need to do is forget that they passed `__return_true` to the filter after a few months and then add a meta box or an option that managed the sensitive data. Thoughts?

fieldmanager.php
```diff
@@ -435,6 +443,100 @@ function fm_trigger_context_action() {
...
+
+		// Quickedit.
+		do_action( "fm_quickedit_{$post_type_endpoints[ $matches[1] ]}" );
```
dlh01: While firing `fm_quickedit_*` addresses the particular issue described earlier, it potentially introduces a separate backwards-compatibility concern.

Is there any precedent in FM for firing multiple context hooks during the same request? Existing FM code using the context actions might already assume that `fm_quickedit_*` will fire on only `edit.php` and not concurrently with `fm_post_*`, for example.

A broader backwards-compatibility concern is that `fm_get_context()` and `fm_calculate_context()` don't work as expected during REST API requests.

php/context/class-fieldmanager-context.php
kjbenk: Correct, but since they are declared within `Fieldmanager_Context_Storable`  it works fine.

php/class-fieldmanager-field.php
kjbenk: My intention was to give developers the ability to expose all fields to the REST API via a single line, and this filter allows them to do that.  I think exposing all fields and then blacklisting the ones that contain sensitive information is a good start.  Overall, it is still up to the developer to make sure they content exposed in the REST API is safe and secure, so I am ok with the way it is.  Though, as always, I would love to hear your thoughts :)

fieldmanager.php
kjbenk: > A broader backwards-compatibility concern is that fm_get_context() and fm_calculate_context() don't work as expected during REST API requests.

Yeah, I think having those two functions handle the action hooks for all REST API events is the best solution.  This work was taken from a branch that Brad worked on and since it worked I just kept it as is.  Maybe we should start to think about how we can leverage `fm_get_context()` and `fm_calculate_context()` to call the proper hooks during a REST request.

fieldmanager.php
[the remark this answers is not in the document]
kjbenk: > Should this also fire fm_post? (Similarly, should fm_term be fired below?)

I am not sure, you might have more incite then me on this one.  However, based on your comment below about [leveraging the existing trigger functions](https://github.com/alleyinteractive/wordpress-fieldmanager/pull/583#discussion_r153648203) I think it might be best to try and incorporate the REST API requests into the trigger functions.

kjbenk:
@dlh01 

> Sorry, let me clarify: A group with 'serialize_data' => false with two children will save two prefixed rows to post meta, but will be returned by the REST API under this branch as one object with two unprefixed keys.

Correct the data will still be exposed within the parent field, though I am not sure that is a bad thing for now.  Do you think it is necessary to make sure the data is exposed in format that is guided by `serialize_data`?

php/class-fieldmanager-options.php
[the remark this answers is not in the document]
dlh01: No, sorry, I meant to just note that the requirements for when `has_built_data` is set to `true` are stricter here than they are in the other method. But I'm not sure whether that's important.

dlh01:
> Correct the data will still be exposed within the parent field, though I am not sure that is a bad thing for now. Do you think it is necessary to make sure the data is exposed in format that is guided by serialize_data?

I'm not sure. Maybe not, given that the data isn't being included with the object as meta.

mboynes:
Finally going to close this. See #404 and #507 for explanations why.

PR #590, opened 2017-02-16
PHPCS: /php/class-fieldmanager-grid.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `/php/class-fieldmanager-grid.php`


mslinnea:
💯

PR #591, opened 2017-02-16
PHPCS: /php/class-fieldmanager-group.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `/php/class-fieldmanager-group.php`

### DIFF
--- a/php/class-fieldmanager-group.php
+++ b/php/class-fieldmanager-group.php
@@ -1,4 +1,9 @@
 <?php
+/**
+ * Class file for Fieldmanager_Group.
+ *
+ * @package Field / Group
+ */
 
 /**
  * Define a groups of fields.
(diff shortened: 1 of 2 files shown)


php/class-fieldmanager-group.php
```diff
@@ -1,4 +1,9 @@
 <?php
+/**
+ * Class file for Fieldmanager_Group.
+ *
+ * @package Field / Group
```
romancandlethoughts: Not sure if this conversation was had at some point, but two things here:
1) According to the [PHPDoc.org docs](https://www.phpdoc.org/docs/latest/references/phpdoc/tags/package.html), `@package` should be the "logical subdivision" of the [structural element](https://www.phpdoc.org/docs/latest/glossary.html#term-structural-element), separated by `\` (similar to namespaces: `@package [level 1]\[level 2]\[etc.]`).
2) There shouldn't be [multiple `@package` tags](https://github.com/alleyinteractive/wordpress-fieldmanager/pull/591/files#diff-65cae30651b720e59f90adc96c1edd35R15) for one structural element.

PR #592, opened 2017-02-16
PHPCS: /php/class-fieldmanager-hidden.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `/php/class-fieldmanager-hidden.php`


mslinnea:
💯

PR #593, opened 2017-02-16
PHPCS: /php/class-fieldmanager-link.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `/php/class-fieldmanager-link.php`

### DIFF
--- a/php/class-fieldmanager-link.php
+++ b/php/class-fieldmanager-link.php
@@ -1,16 +1,20 @@
 <?php
+/**
+ * Class file for Fieldmanager_Link.
+ *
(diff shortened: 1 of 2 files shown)


php/class-fieldmanager-link.php
```diff
@@ -1,21 +1,25 @@
 <?php
+/**
+ * Class file for Fieldmanager_Link.
+ *
+ * @package Field / Link
```
romancandlethoughts: see: https://github.com/alleyinteractive/wordpress-fieldmanager/pull/591/files#r101547999

PR #596, opened 2017-02-16
PHPCS: php/class-fieldmanager-radios.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file

- `php/class-fieldmanager-radios.php`


goodguyry:
🐴

PR #598, opened 2017-02-16
PHPCS: /php/class-fieldmanager-select.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `php/class-fieldmanager-select.php`


goodguyry:
🐴

PR #601, opened 2017-02-17
PHPCS: php/context/class-fieldmanager-context-post.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `php/context/class-fieldmanager-context-post.php`

### DIFF
--- a/php/context/class-fieldmanager-context-post.php
+++ b/php/context/class-fieldmanager-context-post.php
@@ -1,60 +1,73 @@
 <?php
+/**
+ * Class file for Fieldmanager_Context_Post.
+ *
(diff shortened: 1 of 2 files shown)


php/context/class-fieldmanager-context-post.php
```diff
@@ -1,60 +1,73 @@
...
+	 *
 	 * @var string
-	 * Title of meta box
 	 */
 	public $title = '';
 
 	/**
+	 * What post types to render this meta box.
+	 *
 	 * @var string[]
```
davisshaver: same as on another of the PR, is this supposed to be array?

PR #602, opened 2017-02-17
PHPCS: php/context/class-fieldmanager-context-quickedit.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `php/context/class-fieldmanager-context-quickedit.php`

### DIFF
--- a/php/context/class-fieldmanager-context-quickedit.php
+++ b/php/context/class-fieldmanager-context-quickedit.php
@@ -1,66 +1,81 @@
 <?php
+/**
+ * Class file for Fieldmanager_Context_QuickEdit.
+ *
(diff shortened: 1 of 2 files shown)


php/context/class-fieldmanager-context-quickedit.php
```diff
@@ -1,66 +1,81 @@
...
 	 * display in a custom column.
+	 *
+	 * @var callback
 	 */
 	public $column_display_callback = null;
 
 	/**
+	 * What post types to render this Quickedit form.
+	 *
 	 * @var string[]
```
davisshaver: typo here?

PR #610, opened 2017-02-22
PHPCS: php/datasource/class-fieldmanager-datasource-user.php

kjbenk:
# Issue
#271 

# Description
PHPCS for files:

- `php/datasource/class-fieldmanager-datasource-user.php`

### DIFF
--- a/php/datasource/class-fieldmanager-datasource-user.php
+++ b/php/datasource/class-fieldmanager-datasource-user.php
@@ -204,7 +204,24 @@ class Fieldmanager_Datasource_User extends Fieldmanager_Datasource {
 
 		if ( ! empty( $current_values ) ) {
 			foreach ( $current_values as $user_id ) {
-				delete_metadata( 'user', $user_id, $this->reciprocal, $field->data_id );
+				call_user_func(
+					/**
+					 * Allow control over deleting from the user data storage.
+					 * This improves compatibility with WordPress.com.
+					 *
+					 * @see delete_user_meta() for more details about each param.
+					 *
+					 * @param string $function_name The function to call to get user
+					 *                              data. Default is 'delete_user_meta'.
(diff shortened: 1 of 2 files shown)


php/datasource/class-fieldmanager-datasource-user.php
```diff
@@ -114,146 +135,159 @@ public function get_value( $value ) {
...
+	 */
+	public function presave_alter_values( Fieldmanager_Field $field, $values, $current_values ) {
+		if ( 'post' != $field->data_type || ! $this->reciprocal || 'ID' != $this->store_property ) {
 			return $values;
 		}
 
 		if ( ! empty( $current_values ) ) {
 			foreach ( $current_values as $user_id ) {
-				delete_user_meta( $user_id, $this->reciprocal, $field->data_id );
+				delete_metadata( 'user', $user_id, $this->reciprocal, $field->data_id );
```
mboynes: Why the change to `delete_metadata()` (and `add_metadata()` on line 240)?

kjbenk: the linter complains and sends me here - https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#working-with-wp_users-and-user_meta

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/49cb5ce25f4712a2e3d2aa483736372c145f7921/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php#L148-L158

mboynes: that fools the linter, but doesn't fix the problem. we'll need to do something [like we did in the user context](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/context/class-fieldmanager-context-user.php#L126-L148)

PR #614, opened 2017-02-27
PHPCS: templates/textfield.php

kjbenk:
# Issue
#271 

# Description
PHPCS for file:

- `templates/textfield.php`


goodguyry:
🐴

templates/textfield.php
```diff
@@ -1,8 +1,16 @@
+<?php
+/**
+ * Template file for a textfield
```
dlh01: `text field` is two words, I think (e.g. `date field`). 

That said, we already have something of a style for template file headers in `master` in the [`Datepicker`](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/templates/datepicker.php) and [`Checkboxes`](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/templates/options-checkboxes.php) templates.

I don't know whether it matters what the style is, but I think we could try to be consistent.

templates/textfield.php
```diff
@@ -1,8 +1,16 @@
...
+ *
+ * @package Templates
```
dlh01: Similarly, `@package Templates` would cause us to have different `@package`s within the plugin template parts. I used `@package Fieldmanager` in the Datepicker and Checkboxes templates, but I think that was a mistake. Something like `Fieldmanager\Template` might better fit [the existing hierarchy](http://api.fieldmanager.org/).

templates/textfield.php
```diff
@@ -1,8 +1,16 @@
...
 	class="fm-element"
-	type="<?php echo esc_attr( $this->input_type ) ?>"
+	type="<?php echo esc_attr( $this->input_type ); ?>"
 	name="<?php echo esc_attr( $this->get_form_name() ); ?>"
 	id="<?php echo esc_attr( $this->get_element_id() ); ?>"
 	value="<?php echo esc_attr( $value ); ?>"
-	<?php echo $this->get_element_attributes(); ?>
-/>
\ No newline at end of file
+	<?php echo $this->get_element_attributes(); // Escaped in function call WPCS XSS okay. ?>
```
dlh01: Same again: [There is a style for this already in `master`](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/templates/datepicker.php#L18); don't know which is better, but could be consistent.

PR #617, opened 2017-04-26
Cast term_taxonomy_id to string before pushing it to post_meta

david-binda:
Attempting to store an integer in post_meta and option leads to failing new and old value comparison performed before saving the value, which should prevent unnecessary SQL queries. Failing the comparison results in an extra SELECT and UPDATE SQL query, when the UPDATE fails anyway.

More details on the issue can be found here: https://core.trac.wordpress.org/ticket/22192

This commit is casting the integer value of term_id / term_taxonomy_id to string before it's passed to update_post_meta function. This fixes the false-negative result of the mentioned comparison and thus saves 2 SQL queries in case the value does not get changed.

The change introduced here should be compatible with any future change to the update_option and update_metadata functions, at least as proposed in patches in mentioned ticket.


mboynes:
Hey @david-binda, thanks for the PR! Sorry it's taken us a while to get around to digging into it.

I don't know if we can move forward with this one, at least as-is. `Fieldmanager_Datasource_Term::presave()` doesn't only run for top-level, singular fields. It may also run as a sub-field of a group (which ultimately gets serialized into post meta).

I _think_ we can attack this in contexts. If the value being stored is scalar, we could cast the value to a string before passing to `update_post_meta()`.

mboynes:
The change I proposed would likely have to also be made in the user, term, quickedit, and submenu contexts, but that should be confirmed.

david-binda:
Hey Matt! 

> It may also run as a sub-field of a group (which ultimately gets serialized into post meta).

Sounds like a very valid concern.

> I think we can attack this in contexts. If the value being stored is scalar, we could cast the value to a string before passing to update_post_meta().

Sounds like a plan. Let me look into it later this week, unless you have already started.

mboynes:
@david-binda thanks! I did look into making that change and have written all the code in my head at least, so I'll be happy to make those changes :-).

mboynes:
Closing this in favor of #650. Thanks @david-binda!

PR #628, opened 2017-07-14
Fix for tinymce editor not being initialized on richtext fields

mviggiano:
Meta boxes with `Fieldmanager_RichTextArea` fields inside of them that were collapsed on page load did not have their TinyMCE initialized due to not being visible. This PR will add code to check on `postbox-toggle` events whether or not the toggled postbox has such a field, and if so, whether or not it has been initialized. If it has a field that has not been initialized, the code will call `add_rte_to_visible_textareas()` again to add the TinyMCE to the textarea.


js/richtext.js
```diff
@@ -110,10 +110,19 @@
 			if ( 'html' === core_editor_state || 'tinymce' === core_editor_state ) {
 				setUserSetting( 'editor', core_editor_state );
 			}
+		},
+
+		should_readd_rte: function( e, el ) {
```
mviggiano: As far as I'm aware?

js/richtext.js
```diff
@@ -110,10 +110,19 @@
...
+			if ( ! $( el ).hasClass( 'closed' ) && $( el ).find( 'textarea.fm-richtext' ).length ) {
+				if ( ! $( el ).find( 'textarea' ).hasClass( 'fm-tinymce' ) ) {
```
mviggiano: Updated this - thanks for the tip!

mslinnea:
@mboynes @dlh01  I've fixed the merge conflicts here and I think this is ready for 1.4.0.  I tested this solution in a Gutenberg environment and it works there as well.

dlh01:
Thanks @mslinnea!

PR #619, opened 2017-04-28
Updating documentation to reflect the possibility of multiple display_if values...

kkoppenhaver:
...Clarifies alleyinteractive/wordpress-fieldmanager#618

PR #635, opened 2017-08-20
Allow using display-if with an arbitrary selector

dlh01:
For example, to control whether a field displays based on the selected page template:

```
$fm = new \Fieldmanager_TextField( [
	'name' => 'my_template_specific_field',
	'display_if' => [
		'selector' => '#page_template',
		'value' => 'my-page-template.php',
	],
] );
$fm->add_meta_box( 'Page Template-Specific Field', fm_get_context()[1] );
```

The assumed use case with this approach is for controlling field visibility based on only one other element. If `selector` matches multiple elements, the resulting behavior would be unpredictable because the field would toggle based on the last-changed element.

Also brings in the nearby changes from #619.

### DIFF
--- a/php/class-fieldmanager-field.php
+++ b/php/class-fieldmanager-field.php
@@ -203,18 +203,29 @@ abstract class Fieldmanager_Field {
 	public $datasource = Null;
 
 	/**
-	 * @var array[]
 	 * Field name and value on which to display element. Sample:
+	 *
 	 * $element->display_if = array(
-	 *	'src' => 'display-if-src-element',
-	 *	'value' => 'display-if-src-value'
+	 *     'src' => 'display-if-src-element',
+	 *     'value' => 'display-if-src-value',
 	 * );
 	 *
-	 * Multiple values are allowed if comma-separated. Sample:
+	 * Multiple values are allowed if they're comma-separated. Sample:
+	 *
 	 * $element->display_if = array(
-	 *	'src' => 'display-if-src-element',
-	 *	'value' => 'display-if-src-value1,display-if-src-value2'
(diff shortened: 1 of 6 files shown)


php/class-fieldmanager-field.php
```diff
@@ -446,9 +463,29 @@ public function element_markup( $values = array() ) {
...
-			$fm_wrapper_attrs['data-display-src'] = $this->display_if['src'];
+
+			if ( isset( $this->display_if['src'] ) && isset( $this->display_if['selector'] ) ) {
+				throw new \FM_Developer_Exception(
+					esc_html( sprintf(
+						/* translators: 1: 'display-if', 2: `src`, 3: `selector` */
+						__( 'A field cannot have both a %1$s %2$s and a %1$s %3$s', 'fieldmanager' ),
+						'`$display_if`',
+						'`src`',
+						'`selector`'
```
mboynes: Do these 3 strings need to be added via `sprintf()`? Since they aren't dynamic values, is there a reason they can't just be a part of the internationalized string?

dlh01: The idea is that they're not dynamic, but they're also not translatable. I borrowed this pattern from what I've seen in some places in core, e.g. https://github.com/WordPress/WordPress/blob/master/wp-admin/network.php#L40-L49. I assume the reasoning is that it makes the remaining string easier to translate, but I can't point to any documentation about it.

dlh01: Another example, hot off the press: https://core.trac.wordpress.org/ticket/42161.

mboynes: That makes good sense, thanks for clarifying!

PR #641, opened 2017-10-02
Test against PHP 7.1 in Travis

dlh01:
WordPress core has not allowed its tests to fail on PHP 7.1 [since December 2016](https://core.trac.wordpress.org/changeset/39424). Accordingly, this requires Fieldmanager to run against 7.1 per the reasoning in #462.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: php
 matrix:
   include:
     - php: 5.3
-      env: WP_VERSION=4.4.4
+      env: WP_VERSION=4.4.11
       dist: precise
     - php: 5.3
       env: WP_VERSION=latest
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,19 @@ matrix:
       dist: precise
     - php: 5.6
       env: WP_VERSION=latest
-    - php: 7.1
-      env: WP_VERSION=latest
-    - php: 5.6
-      env: WP_VERSION=nightly
     - php: 5.6
       env: WP_TRAVISCI=phpcs
+    - php: 7.0
+      env: WP_VERSION=latest
+    - php: 7.0
+      env: WP_VERSION=nightly
+    - php: 7.1
+      env: WP_VERSION=latest
+    - php: 7.2
+      env: WP_VERSION=latest
   fast_finish: true
+  allow_failures:
+    - php: 7.2
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
       dist: precise
     - php: 5.6
       env: WP_VERSION=latest
-    - php: 7.0
+    - php: 7.1
       env: WP_VERSION=latest
     - php: 5.6
       env: WP_VERSION=nightly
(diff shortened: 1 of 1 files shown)


dlh01:
The circumstances of the failing test looks very similar to me to this bug reported against PHP: https://bugs.php.net/bug.php?id=75148.

I'm a little confused about whether the warning is expected. The PHP 7.1 migration guide says ["call_user_func() will now always generate a warning upon calls to functions that expect references as arguments."](http://php.net/manual/en/migration71.incompatible.php#migration71.incompatible.call_user_func-with-ref-args), but `call_user_func_array()` is generating this warning, in `WP_Hook::apply_filters()`. 

Core also addressed a similar warning by switching to `call_user_func_array()` in https://core.trac.wordpress.org/changeset/38235, which leads me to think it should be safe to use with functions that expect references.

So, I think I'm missing something, and I'm not sure what the best course is to address the failure in this PR. 

The method in Fieldmanager that expects the reference, `Fieldmanager_Datasource_Post::title_like()`, doesn't even use the `&$wp_query` parameter, so one option could be to just remove the parameter:

```diff
diff --git a/php/datasource/class-fieldmanager-datasource-post.php b/php/datasource/class-fieldmanager-datasource-post.php
index 4511ce5..47d9d0d 100644
--- a/php/datasource/class-fieldmanager-datasource-post.php
+++ b/php/datasource/class-fieldmanager-datasource-post.php
@@ -175,7 +175,7 @@ class Fieldmanager_Datasource_Post extends Fieldmanager_Datasource {
     /**
      * Perform a LIKE search on post_title, since 's' in WP_Query is too fuzzy when trying to autocomplete a title
      */
-    public function title_like( $where, &$wp_query ) {
+    public function title_like( $where ) {
         global $wpdb;
         if ( method_exists( $wpdb, 'esc_like' ) ) {
             $like = esc_sql( $wpdb->esc_like( $this->_fragment ) );
```

mboynes:
The solution should be to simply remove the `&`. Objects are always passed by reference, so it doesn't change anything. I ran into [the same problem on ES_WP_Query](https://github.com/alleyinteractive/es-wp-query/pull/58/commits/f7e86eb2aa0d48f28bd427057ceae84ca3b77295).

.travis.yml
```diff
@@ -12,7 +12,7 @@ matrix:
       dist: precise
     - php: 5.6
       env: WP_VERSION=latest
-    - php: 7.0
+    - php: 7.1
```
mboynes: I think one of the `5.6` builds should probably be switched to `7.0` so that gets run. Also might be worth considering adding a 7.2 build (which is allowed to fail, since 7.2 is pre-release)

dlh01:
As of two days ago, WordPress also [no longer allows its tests to fail on PHP 7.2](https://core.trac.wordpress.org/changeset/41748). However, as @mboynes noted, PHP 7.2 isn't officially released. While we probably don't need to test against each of 7.0, 7.1, and 7.2, we don't want to drop the latest stable release from the Travis matrix, either.

So, the logic in the Travis matrix is now slightly modified from that of #462: "Test against the highest version of PHP that core tests against, but require tests to pass against the highest stable version of PHP." When PHP 7.2 is officially released, we should be able to remove 7.2 from `allow_failures` and remove 7.1 from the tests.

PR #647, opened 2017-10-09
Allow specifying an array of "Add Another" labels

### LINKED ISSUE #630: Allow specifying array of "Add More" labels
`Fieldmanager_Field::add_more_label` currently accepts a string. It might be useful for it to also accept an array of labels for different states of repeatable-ness. Some possibilities:

* `add_first`, when no items exist.
* `add_more`, when at least one item exists.
* `limit_reached`, when no more items can be added. See #629. The button could also be disabled in this case.
* `minimum_count_reached`, when no more items can be removed. Perhaps out-of-scope for this issue, though, because the "Remove" link isn't the same as the button. See #582.

dlh01:
Fixes #630.

### DIFF
--- a/php/class-fieldmanager-field.php
+++ b/php/class-fieldmanager-field.php
@@ -62,7 +62,20 @@ abstract class Fieldmanager_Field {
 	/**
 	 * Labels associated with the "Add Another" button.
 	 *
-	 * @since 1.2.0 Converted from a single label to an array of labels.
+	 * Accepted keys for the `add_more_label` array:
+	 *
+	 * - `add_first`, used as the button text when no items have been added.
+	 *   Default is 'Add field' or 'Add group'.
+	 *
+	 * - `add_another`, used as the button text when at least one item has been
+	 *   added. Default is 'Add field' or 'Add group'.
+	 *
+	 * - `limit_reached`, displayed alongside the button when no more items can
+	 *   be added. Default is none for better backwards-compatibility with
+	 *   existing fields that might already specify the limit in the field's
+	 *   label or description.
+	 *
+	 * @since 1.4.0 Converted from a single label to an array of labels.
 	 *
 	 * @see Fieldmanager_Field::add_another().
 	 *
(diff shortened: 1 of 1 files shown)


dlh01:
@stevenslack I would love to hear your thoughts about the UI and UX of these new labels, particularly given your work in #468.

Here's a small field I used while testing the new labels:

```php
<?php
add_action( 'fm_post', function () {
	$fm = new \Fieldmanager_TextField( [
		'name' => 'add_another',
		'limit' => 4,
		'minimum_count' => 2,
		'add_more_label' => [
			'add_first' => 'Add first',
			'limit_reached' => 'Limit reached',
		],
	] );
	$fm->add_meta_box( 'Add Another', fm_get_context()[1] );
} );
```

css/fieldmanager.css
```diff
@@ -124,6 +124,15 @@ div.fm-group-label-wrapper {
...
 }
 
+.fm-add-another-limit-reached-label {
+	/* These sizes correspond to default button sizes in core. */
+	display: inline-block;
+	font-size: 13px;
+	height: 28px;
+	line-height: 26px;
+	margin: 0 10px;
+}
```
stevenslack, line 134: I wonder if we could also utilize core classes to output the desired button appearance rather than using our own class. This is especially true for `height`, `font-size`, `line-height` etc.

stevenslack (approved):
@dlh01 I tested this field locally and I do think it is a nice improvement to the user experience. I had one thought about how we are styling our buttons as it relates to core button styles. We may want to consider leveraging core classes to get the style we want. This could however prove difficult if trying to provide legacy support for different WP versions, although I am unaware of any implications this may or may not provide. The current approach provided in this PR is no blocker, just a thought.

dlh01:
Totally agree, @stevenslack. I think somewhere out there is a ticket or a `@todo` or something to look into doing that. (If not, let's make one.) If memory serves, the core button classes are maybe newer than Fieldmanager but quite stable at this point. I don't anticipate there would be backwards compatibility concerns.

dlh01:
The JS implementation in this PR falls apart with repeatable fields in group children. I'm going to bump it out of the 1.4.0 milestone, but anyone feeling motivated to remediate the PR sooner is welcome to.

PR #650, opened 2017-10-10
Cast scalar meta values as string

mboynes:
Replaces #617, props @david-binda

### DIFF
--- a/php/context/class-fieldmanager-context-post.php
+++ b/php/context/class-fieldmanager-context-post.php
@@ -310,14 +310,7 @@ class Fieldmanager_Context_Post extends Fieldmanager_Context_Storable {
 	 *                                Default empty.
 	 */
 	protected function update_data( $post_id, $meta_key, $meta_value, $data_prev_value = '' ) {
-		/*
-		 * Meta is always stored as a string, so if this is a scalar value, cast
-		 * it as a string to ensure that `update_metadata()` is able to correctly
-		 * compare the current value against the previous value.
-		 */
-		if ( is_scalar( $meta_value ) && ! is_string( $meta_value ) ) {
-			$meta_value = strval( $meta_value );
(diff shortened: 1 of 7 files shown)


php/context/class-fieldmanager-context-post.php
```diff
@@ -310,6 +310,12 @@ protected function add_data( $post_id, $meta_key, $meta_value, $unique = false )
 	 *                                Default empty.
 	 */
 	protected function update_data( $post_id, $meta_key, $meta_value, $data_prev_value = '' ) {
+		// Meta is always stored as a string, so if this is a scalar value, cast
+		// it as a string to ensure that `update_metadata()` is able to correctly
+		// compare the current value against the previous value.
```
dlh01: Perhaps should follow docs standards for inline comments: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#5-2-multi-line-comments (applies throughout)?

php/context/class-fieldmanager-context-submenu.php
```diff
@@ -239,6 +239,12 @@ protected function add_data( $data_id, $option_name, $option_value, $unique = fa
 	 * @return bool Option updated successfully.
 	 */
 	protected function update_data( $data_id, $option_name, $option_value, $option_prev_value = '' ) {
+		// Meta is always stored as a string, so if this is a scalar value, cast
+		// it as a string to ensure that `update_metadata()` is able to correctly
```
dlh01: Should this be `*_option()` instead of `_metadata`, and the earlier reference to "Meta is" be "Options are"?

mboynes:
Thanks for the suggestion @dlh01. Not sure why I didn't consider that, it seems so obvious in retrospect. That also made it more testable, of course.

PR #651, opened 2017-10-11
Increase minimum supported WP version to 4.7

mboynes:


dlh01:
If PHP5 isn't officially supported, then you could argue a Travis failure on PHP5 wouldn't in itself be a reason to block a merge. Should the PHP5 tests therefore be in `allow_failures`?

mboynes:
Nah, I still want to _try_ to support it, I just don't want to make any guarantees.

Issue #625, opened 2017-06-26
Don't submit prototype fields for repeatables


mboynes:
Related to #624, large forms can lead to data loss depending on the server configuration. When a group or field is repeatable, Fieldmanager outputs a "prototype" for that repeatable element, then new elements clone it. These hidden "prototype" elements are still part of the form and get submitted with the form.

Fieldmanager should either remove them prior to submission via JS, or (preferably) the prototypes should never be a part of the DOM in the first place. A _relatively_ straightforward solution would be to leverage Underscore templates and change the repeatable functions to use the Underscore templates to build new fields.

PR #658, opened 2017-10-12
Potential fix for fm_renumber and radios bug

### LINKED ISSUE #548: Radio values disappear upon dragging repeating sortable field group
I have a repeating sortable field group containing a set of radio fields. When I drag this group to re-order, the field values sometimes disappear. But when I save the post, the correct values reappear.

mboynes:
A proposed solution to the bug reported in #548, this code modifies `fm_renumber()` by adding a temporary and unique prefix to all modified field names during the renumbering process, then removes that prefix after renumbering is done.

I'm not sold on this approach, it feels like a lot of overhead.

Closes #548


dlh01:
The `_.uniqueId()` function in Underscore is intended for "temporary DOM ids," which seems similar to what you're going for? http://underscorejs.org/docs/underscore.html#section-164

While that doesn't really reduce the "overhead" of the change (and might even increase the overhead because we need to make sure Underscore loads before `fieldmanager.js`, although #625 might have the same effect), it might allow us to offload some of the work to an existing library, and it might at least "feel" cleaner.

mgburns:
Confirming that this branch does fix the issue for us! 🥇

montchr:
It works for us too! Thank you.

kaitlinbolling:
This fix worked for me! Can we get this merged?

dlh01:
While it might be more overhead than ideal in theory (or might not), in practice there are many testimonials to it working, and no new approaches have been presented, so I think we can move ahead. We can always improve its performance in future PRs.

dlh01 (approved):
🌳

PR #659, opened 2017-10-13
Fix issues with only_save_to_taxonomy

srtfisher:
When `$only_save_to_taxonomy` is set to true, some errors can occur. `get_the_terms()` can return false when creating a new post. This (when passed to `usort()` throws an error since it's not an array.

We should not use the return value of  `usort() ` as it sorts the array by reference.


srtfisher:
@mboynes and we're green!

PR #664, opened 2017-10-24
Allow limiting to top-level terms, for which Core uses parent of `0`

ethitter:


ethitter:
Good call! Test added.

PR #673, opened 2017-11-26
Move the array `[]` appended to form names to `Fieldmanager_Options`

### LINKED ISSUE #622: Incorrect rule key for Fieldmanager_Checkboxes validation_rules
When adding `validation_rules` to `Fieldmanager_Checkboxes` the `rules` key in the `jQuery.validate` initialization object is the "singular" name for the field when it should be added as an array:

```js
$( '#post' ).validate( {
  ...
  rules: {
    "checkboxesfield": { // <- This doesn't work
      required: true
    },
    ...
```

It should be `checkboxesfield[]`, not `checkboxesfield`:

```js
$( '#post' ).validate( {
  ...
  rules: {
    "checkboxesfield[]": { // <- This works!
      required: true
    },
    ...
```

dlh01:
Fixes #622. The `[]` had been appended manually to `Fieldmanager_Select` and `Fieldmanager_Checkboxes` form names. But, `Fieldmanager_Util_Validation` relies only on `Fieldmanager_Field::get_form_name()` for determining the name, which wouldn't include `[]`.

With this change, the `[]` in multi-selects and checkboxes is generated by overriding `Fieldmanager_Field::get_form_name()` and checking the `$multiple` property, which ensures `[]` is also returned to other callers like `Fieldmanager_Util_Validation`.

However, this change assumes that any field with `'multiple' => true` should in fact be represented as an array in its form name, which happens to be true for the bundled fields but might not be true for custom classes or future bundled fields. Those fields would need to override `get_form_name()` to ensure it behaved as expected.

An alternative approach with fewer backwards-compatibility concerns might be to override `get_form_name()` on `Fieldmanager_Select` and `Fieldmanager_Checkboxes`. The primary downside would be that `get_form_name()` would need to be similarly overridden on future subclasses of `Fieldmanager_Options` that were arrays.

### DIFF
--- a/php/class-fieldmanager-options.php
+++ b/php/class-fieldmanager-options.php
@@ -200,6 +200,24 @@ abstract class Fieldmanager_Options extends Fieldmanager_Field {
 	}
 
 	/**
+	 * Override the default HTML form name to allow denoting arrays.
+	 *
+	 * @since 1.3.0
+	 *
+	 * @return string Form name.
+	 */
+	public function get_form_name( $multiple = '' ) {
+		// Don't override anything passed to subclasses before this was implemented.
+		if ( ! $multiple ) {
+			if ( $this->multiple ) {
+				$multiple = '[]';
+			}
+		}
+
+		return parent::get_form_name( $multiple );
+	}
+
+	/**
 	 * Helper for output functions to toggle a selected options.
 	 *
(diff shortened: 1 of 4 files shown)


juliobranha (approved):
🐐 
I've tried this on local for HBG project and worked as expected :)

Issue #627, opened 2017-06-27
Refresh Nonce Values After Reauthenticating To Prevent Data Loss


kingkool68:
## Problem
When an authenticated session expires and a user logs in again the FieldManager nonce fields should refresh as well. The way things work now is when you reauthenticate everything seems fine so a user may continue adding content. When the post is eventually saved the user is presented with a `wp_die()` screen letting them know something is wrong (because behind the scenes the nonce verification failed). All of their previous work is lost after this happens.

This has happened to our editorial team a number of times and it is painful to hear how they lose work whenever it happens. 

## Suggested Solution
WordPress provides a filter, `wp_refresh_nonces`, for sending new nonces to the editor after reauthenticating. You can see an example of how this is used in the `wp_refresh_post_nonces()` function in `/wp-admin/includes/misc.php`.  You provide a key => value pair with the key being the ID attribute of the nonce and the value being the new nonce value. WordPress should handle doing the replacement for you in `post.js`. 

## Steps to Reproduce
 - Create a new post with one or more FieldManager fields
 - Save the post as a draft
 - Make more content changes
 - Delete the domain cookies to force WordPress to reauthenticate via the Heartbeat API
 - Reauthenticate
 - Save the post and be presented with an error screen and all your changes are gone

mboynes:
Thanks for the excellent writeup!

We'll need a solution that works across all contexts. Hopefully core makes that easy, but I've not looked into it. If anyone has time to take this on, please do. Otherwise, I might not be able to get to it until next month or even later.

dlh01:
We expect that #677 will address this issue, excepting the limitations in the block editor and non-post contexts described in the PR. If not, please reopen with the details!

PR #677, opened 2017-12-04
Refresh nonces with heartbeat responses

dlh01:
See #627.

The implementation in this PR makes a filter for `wp_refresh_nonces` available to all contexts by placing it on the base `Fieldmanager_Context` class. However, to add the filter, custom contexts would need to add calls to `parent::__construct()` or call `add_filter()` manually.

It also appears that core does not refresh its own nonces on heartbeat requests outside of the edit-post screen. For example, unless I was doing something wrong, if you mimic the steps in #627 on an edit-term screen, attempting to save the term after reauthenticating will still fail. See https://core.trac.wordpress.org/ticket/29312.


mslinnea:
Hey @dlh01,
I tested this following the steps outlined in #627 
This works successfully in WordPress 5.5.2 and WordPress 5.6 for the classic post editor.  On the Gutenberg post editor, I was not able to complete the testing steps because clicking on the Gutenberg "save draft" link fails after reauthenticating.  
I also updated this PR with master and resolved the merge conflicts.
Looks good to me!

dlh01:
Thanks @mslinnea! As we discussed outside of GitHub, the source of the failing nonce appears to be WordPress core (and we'll work on filing a Trac ticket for it soon). 

I checked the `POST` request for saving meta boxes in the block editor, and I confirmed that the refreshed nonces for Fieldmanager are still present. So, once the core bug is fixed, I would expect that this PR will have the desired result in the block editor as well as the classic editor.

Would you mind giving the PR a formal 👍 if everything else seems OK?

mslinnea (approved):
Looks good! 🐴

PR #678, opened 2017-12-07
Fix a bug where checkboxes trigger displayif

mboynes:
Also includes a version bump; this will push out as 1.2.1.

### DIFF
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -206,8 +206,9 @@ $( document ).ready( function () {
 		var src = $( this ).data( 'display-src' );
 		var values = getCompareValues( this );
 		// Wrapper divs sometimes receive .fm-element, but don't use them as
-		// triggers. Also don't use autocomplete inputs as triggers, because the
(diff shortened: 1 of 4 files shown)


js/fieldmanager.js
```diff
@@ -208,7 +208,7 @@ $( document ).ready( function () {
 		// Wrapper divs sometimes receive .fm-element, but don't use them as
 		// triggers. Also don't use autocomplete inputs as triggers, because the
 		// value is in their sibling hidden fields (which this still matches).
-		var trigger = $( this ).siblings( '.fm-' + src + '-wrapper' ).find( '.fm-element' ).not( 'div, .fm-autocomplete' );
+		var trigger = $( this ).siblings( '.fm-' + src + '-wrapper' ).find( '.fm-element' ).not( 'div, .fm-autocomplete, .fm-checkbox-hidden' );
```
dlh01: Might be worth amending the comment above for this additional case.

PR #681, opened 2017-12-18
Support `display_if` for page templates

### LINKED ISSUE #414: Add "display_if" integration for page templates
Title pretty much covers it: show/hide fields based on the page template selected.

ethitter:
Page templates are a special field in Core, and rather than requiring hidden fields to support conditioning a field on them, FM should handle them accordingly. 

By specifying the `page_template` key instead of the `src` and `value` keys, this will toggle fields based on the chosen page template.

Fixes #414

### DIFF
--- a/tests/php/test_fieldmanager_field.php
+++ b/tests/php/test_fieldmanager_field.php
@@ -1078,4 +1078,46 @@ class Fieldmanager_Field_Test extends WP_UnitTestCase {
 		$fm->element_markup( rand_str() );
 		$this->assertSame( 4, $ma->get_call_count(), 'Missing calls to element-markup filters' );
 	}
+
+	public function test_display_if_markup_for_page_template() {
+		$name     = rand_str();
+		$template = 'toaster.php';
+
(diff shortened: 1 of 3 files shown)


tests/php/test_fieldmanager_field.php
```diff
@@ -1078,4 +1078,46 @@ public function test_element_markup_filters() {
 		$fm->element_markup( rand_str() );
 		$this->assertSame( 4, $ma->get_call_count(), 'Missing calls to element-markup filters' );
 	}
+
+	public function test_display_if_markup_for_page_template() {
+		$name     = rand_str();
+		$template = 'toaster.php';
```
kevinfodness: I'm sure there's a good story behind this.

ethitter: 😄 

I wish it was more than this being a word unlikely to come up in a non-testing context. Instead of `foo` and `bar`, I use `waffle iron` and `toaster pastry`, for reasons I can no longer recall; probably something about being self-taught and learning programming when I was a hungry teenager. 😂

PR #684, opened 2018-01-08
Display If checkbox bug

kjbenk:
Add `.fm-checkbox-hidden` to exclude fields to trigger display if


kjbenk:
Closing in favor of #678

PR #689, opened 2018-01-31
Reinit $messages as an array if validation_rules is not a string

terrarum:
I have a field with these validation rules:

```
'validation_rules'   => ['required' => true, 'min' => 1]
```

and I wanted to customise the error message for the `min` rule not being met, which I tried with

```
'validation_messages' => [
  'min' => 'You must select an image'
]
```

which did not work, and gave me this warning:

```
Warning: Illegal string offset 'min' in /vagrant/wordpress/plugins/fieldmanager/php/util/class-fieldmanager-util-validation.php on line 172
```

Investigating `class-fieldmanager-util-validation.php` showed that `$messages` is initialised as a string, but if `$fm->validation_rules` is not a string the code tries to treat `$messages` as an associative array.

### DIFF
--- a/php/util/class-fieldmanager-util-validation.php
+++ b/php/util/class-fieldmanager-util-validation.php
@@ -142,7 +142,7 @@ class Fieldmanager_Util_Validation {
 
 		// Determine if the rules are a string or an array and ensure they are valid.
 		// Also aggregate any messages that were set for the rules, ignoring any messages that don't match a rule.
-		$messages = '';
+		$messages = array();
 		if ( ! is_array( $fm->validation_rules ) ) {
 			// If a string, the only acceptable value is "required".
 			if ( ! is_string( $fm->validation_rules ) || 'required' !== $fm->validation_rules ) {
@@ -159,7 +159,6 @@ class Fieldmanager_Util_Validation {
 				$messages['required'] = $fm->validation_messages;
 			}
 		} else {
-			$messages = array();
 			// Verify each rule defined in the array is valid and also check for any messages that were defined for each.
 			foreach ( $fm->validation_rules as $validation_key => $validation_rule ) {
 				if ( ! in_array( $validation_key, $this->valid_rules ) ) {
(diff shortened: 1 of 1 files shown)


dlh01:
I'm not seeing any place in that method in which `$messages` is used as a string. If it's not, then perhaps `$messages` could just be initialized as an array to begin with.

terrarum:
You are correct! Changed the original $messages to an array, removed my line, it worked.

terrarum:
Hi @dlh01, not sure what the protocol is for this but is there any chance this could be reviewed and merged soon?

dlh01:
@terrarum I'm not sure. It's a tiny change, but I think it might benefit from a second look by someone who's more familiar with the validation functionality than I am.

terrarum:
@dlh01 fair enough, thank you!

terrarum:
@kjbenk @bcampeau Hello, according to git you wrote most of class-fieldmanager-util-validation.php with some small tweaks from @mboynes, is there any chance one of you would be able to review my PR? Thank you!

simonrcodrington:
When will this issue be addressed? 

I've ran into this exact same issue and after a while debugging why jQuery Validator functionality was not working. I ended up changing the `$messages` variables to an array and everything works as expected.

The documentation says that both a string or associative array can be used, but most people I feel will opt into using an array as you can easily specify multiple rules and messages e.g 

`        $textarea = new \Fieldmanager_TextArea(
            'Article Abstract',
            [
                'name'                => 'my_textarea',
                'attributes'          => [
                    'rows'        => 5,
                    'maxlength'   => 225
                ],
                'validation_rules'    => ['digits' => true, 'required' => true],
                'validation_messages' => ['digits'   => 'Only digits are allowed in here',
                                          'required' => 'This field is required, please complete it'
                ]
            ]
        );`

Without this bugfix the validation wont work.

digiguru:
Given a number of people have approved this change (but none with write permission), may I politely ask someone with write permission (@dlh01) to consider a merge now?

Or is there something else you need for extra confidence?

PR #697, opened 2018-03-12
Custom template directory

damonsharp:
Although the form element can be rendered by overriding `Fieldmanager_Field::form_element()`, it would be nice to just provide a custom directory for templates as the second parameter of `fieldmanager_get_template()` for reusability.

For example, I'm building custom admin interfaces that need buttons with specific data-attributes, and it would be nice to just set the options, attributes, etc., and call the `button.php` template, which is being placed outside the default Fieldmanager `/templates` directory, like so...

```$this->template = fieldmanager_get_template( 'button', 'some_custom_dir' );```

This modifies the `fieldmanager_get_template()` function with a second parameter to allow for this.

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -127,9 +127,9 @@ function fieldmanager_get_baseurl() {
 /**
  * Get the path to a field template.
  *
- * @param string $tpl_dir The template directory.
  * @param string $tpl_slug The name of a template file inside the template
  *     directory, excluding ".php".
+ * @param string $tpl_dir The template directory.
  * @return string The template path, or the path to "textfield.php" if the
  *     requested template is not found.
  */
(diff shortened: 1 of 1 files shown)


dlh01:
Hey @damonsharp!

At first glance, this seems like it could potentially be achieved a little more naturally by making the path filterable.

I'm having a little trouble following the scenario as described, though. If this would address the issue:

`$this->template = fieldmanager_get_template( 'button', 'some_custom_dir' );`

then I'm not sure why it couldn't already be `$this->template = '/completely/custom/path.php;` 

Are there cases in which the `template` property can't be set? Or would it be easier to not have to set `template` at all?

damonsharp:
Cleanup/closing outstanding global PRs.

Issue #701, opened 2018-04-04
Saving empty multi-select field in group causes PHP Warnings/Notices


mboynes:
## Steps to reproduce:

1. Add the following code:

	```php
	add_action( 'fm_post_post', function() {
		$fm = new Fieldmanager_Group( [
			'name' => 'group',
			'children' => [
				'field' => new Fieldmanager_Select( [
					'multiple' => true,
					'options' => [ 'One', 'Two' ],
				] ),
			],
		] );
		$fm->add_meta_box( 'Testing', 'post' );
	} );
	```

2. Add a post, don't select an option/deselect one if one is selected, and save the post.

This results in the following warnings and notices (with sample stack trace):

```
( ! ) Warning: Illegal string offset 'field' in /var/www/wordpress/wp-content/plugins/wordpress-fieldmanager/php/class-fieldmanager-group.php on line 359
--


1 | 0.0026 | 424224 | {main}( ) | .../post.php:0
2 | 0.2975 | 4330736 | edit_post( ??? ) | .../post.php:194
3 | 0.3057 | 4386312 | wp_update_post( array(67), ??? ) | .../post.php:377
4 | 0.3072 | 4401864 | wp_insert_post( array(81), false ) | .../post.php:3654
5 | 0.3320 | 4741056 | do_action( string(9), long, object(WP_Post)[1581], true ) | .../post.php:3581
6 | 0.3320 | 4741432 | WP_Hook->do_action( array(3) ) | .../plugin.php:453
7 | 0.3320 | 4741432 | WP_Hook->apply_filters( string(0), array(3) ) | .../class-wp-hook.php:310
8 | 0.3321 | 4743120 | Fieldmanager_Context_Post->delegate_save_post( long ) | .../class-wp-hook.php:288
9 | 0.3321 | 4743120 | Fieldmanager_Context_Post->save_fields_for_post( long ) | .../class-fieldmanager-context-post.php:165
10 | 0.3323 | 4744616 | Fieldmanager_Context_Post->save_to_post_meta( long, ??? ) | .../class-fieldmanager-context-post.php:220
11 | 0.3323 | 4744616 | Fieldmanager_Context_Post->save( null ) | .../class-fieldmanager-context-post.php:255
12 | 0.3323 | 4744616 | Fieldmanager_Context_Post->save_field( object(Fieldmanager_Group)[932], null, long ) | .../class-fieldmanager-context-storable.php:52
[...]

Issue #702, opened 2018-04-04
Saving empty multi-select field in group using term datasource causes PHP Fatal error


mboynes:
## Steps to reproduce:

1. Add the following code:

	```php
	add_action( 'fm_post_post', function() {
		$fm = new \Fieldmanager_Group( [
			'name' => 'group',
			'children' => [
				'field' => new \Fieldmanager_Select( [
					'remove_default_meta_boxes' => true,
					'multiple' => true,
					'datasource' => new \Fieldmanager_Datasource_Term( [
						'taxonomy' => 'category',
						'only_save_to_taxonomy' => true,
					] ),
				] ),
			],
		] );
		$fm->add_meta_box( 'Testing', 'post' );
	} );
	```

2. Add a post, don't select a category/deselect one if one is selected, and save the post.

This results in the following fatal error (with sample stack trace):

```
( ! ) Fatal error: Uncaught Error: Cannot unset string offsets in /var/www/wordpress/wp-content/plugins/wordpress-fieldmanager/php/class-fieldmanager-group.php on line 381
--



1 | 0.0019 | 423744 | {main}( ) | .../post.php:0
2 | 0.2893 | 4337392 | edit_post( ??? ) | .../post.php:194
3 | 0.3053 | 4409136 | wp_update_post( array(63), ??? ) | .../post.php:377
4 | 0.3075 | 4424976 | wp_insert_post( array(78), false ) | .../post.php:3654
5 | 0.3475 | 4783280 | do_action( string(9), long, object(WP_Post)[1615], true ) | .../post.php:3581
6 | 0.3475 | 4783656 | WP_Hook->do_action( array(3) ) | .../plugin.php:453
7 | 0.3475 | 4783656 | WP_Hook->apply_filters( string(0), array(3) ) | .../class-wp-hook.php:310
8 | 0.3478 | 4785344 | Fieldmanager_Context_Post->delegate_save_post( long ) | .../class-wp-hook.php:288
9 | 0.3478 | 4785344 | Fieldmanager_Context_Post->save_fields_for_post( long ) | .../class-fieldmanager-context-post.php:165
10 | 0.3480 | 4785560 | Fieldmanager_Context_Post->save_to_post_meta( long, ??? ) | .../class-fieldmanager-context-post.php:220
11 | 0.3480 | 4785560 | Fieldmanager_Context_Post->save( null ) | .../class-fieldmanager-context-post.php:255
[...]

PR #705, opened 2018-04-04
handle php 7.2 error with saving empty groups

mslinnea:
It looks like this fixes issues #701 and #702 

I think the cause of these errors is actually on line 364 of this file, which assumes that $values is an array.  But $values is actually an empty string when the group is entirely empty.

Although this bug was happening with `multiple => true` it can also happen when `first_empty => true` so I don't think it's specific to the multiple attribute.


php/class-fieldmanager-group.php
```diff
@@ -350,6 +350,11 @@ public function presave( $values, $current_values = array() ) {
 			}
 		}
 
+		// Handle empty groups.
```
dlh01: I wonder whether it's worth adding some details to this comment, or even examples of the scenarios in which it's needed. "Empty groups" could be confused with the `group_is_empty` property, which I don't think (?) is intended.

php/class-fieldmanager-group.php
```diff
@@ -350,6 +350,11 @@ public function presave( $values, $current_values = array() ) {
...
+		if ( empty( $values ) ) {
+			$values = array();
```
dlh01: L343 above also assumes `$values` is an array. Is there ever a case where `$values` could be a legal non-empty non-array? If not, should an exception be thrown before L343 if one is passed?

php/class-fieldmanager-group.php
mslinnea: Good point, this is actually for handling groups that have a multiple select in them, and possible other scenarios -- I'm going to try to flush out what other scenarios that would include.

PR #708, opened 2018-05-02
$messages is used as an array, so it should be created as one.

discoinfiltrator:
A coworker noticed this bug. $messages is being created as a string, but used like:

```
$messages['required'] = $fm->validation_messages;
```


discoinfiltrator:
Dupe of https://github.com/alleyinteractive/wordpress-fieldmanager/pull/689

PR #717, opened 2018-07-16
DIVE-131 Add Gutenberg support for dynamic fields

### LINKED ISSUE #710: Autocomplete fields don't work in side context in Gutenberg
In the Gutenberg editor, autocomplete fields don't work when added to the `'side'` context. I suspect the reason may be to do with the way that the autocomplete events are attached to elements when the DOM is ready. Something about the `'side'` fields tucked in the "Extended Settings" drawer in the current Gutenberg experience seems to prevent this from happening.

# What should happen
Typing in an autocomplete field in the side context should fire an event to query the datasource with the string data that has been entered.

# What happens
When typing in an autocomplete field, no event seems to fire and no query is attempted.

### LINKED ISSUE #713: Fieldmanager_Datepicker field JavaScript does not work with Gutenberg enabled
We are testing an existing site with Gutenberg, and I discovered that the `Fieldmanager_Datepicker` field doesn't work with Gutenberg enabled.

## Expected functionality
Click the field, and a jQuery UI Datepicker calendar pops up to select the date.

## Actual functionality
Click the field, and no calendar pops up. You can still type the date into field, but there is no calendar available to pick the date.

kevinfodness:
* Modifies selectors that instantiate dynamic functionality (e.g., datepickers, color pickers, autocomplete, etc) to explicitly check whether they are within a proto block rather than performing a naïve check based on `:visible`.
* Adds polyfills for `Element.matches` and `Element.closest` to efficiently enable the parent check.
* Adds `package.lock` for Node modules.

Fixes #713 
Fixes #710


js/fieldmanager.js
```diff
@@ -1,5 +1,34 @@
 var fm = {};
 
+/**
+ * Polyfill Element.matches, if necessary.
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill
```
goodguyry: The link should be `https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Polyfill`

kevinfodness: This specific `matches` polyfill was taken from the `closest` polyfill page, which is slightly different from the one present on the `matches` polyfill page you linked above. I'm only including it because `closest` depends on it, so I'm pulling both polyfills from the same location. Good catch, though!

psorensen:
@kevinfodness Just tested this locally and it works great to fix the autocomplete bug. Any idea if/when this will be merged?

kevinfodness:
Closing in favor of #757

Issue #723, opened 2018-09-13
Date displays wrong if date is before 1970s


ravatparmar:
FieldManager-Datepicker shows the wrong date if the date is before 1970, 

The issue is when we have a before the 1970s date at that time timestamp will have a negative value and the Fieldmanager_Datepicker convert the negative value to the positive value so it displays the wrong date.

cc: @danielbachhuber

ravatparmar:
PR: https://github.com/alleyinteractive/wordpress-fieldmanager/pull/724

danielbachhuber:
@ravatparmar Can you provide a test case too?

ravatparmar:
@danielbachhuber Added one test case, and please let me know if we need more test cases or I missed anything.

danielbachhuber:
👍 Thanks @ravatparmar

dlh01:
Fixed in #724.

PR #724, opened 2018-09-13
Fix date displaying wrong if it's before 1970s

ravatparmar:
Issue : https://github.com/alleyinteractive/wordpress-fieldmanager/issues/723

Removed absint() so from Fieldmanager_Datepicker's form_element won't convert the negative timestamp to positive.

### DIFF
--- a/php/class-fieldmanager-datepicker.php
+++ b/php/class-fieldmanager-datepicker.php
@@ -90,6 +90,7 @@ class Fieldmanager_Datepicker extends Fieldmanager_Field {
 	 * @return string HTML for the element.
 	 */
 	public function form_element( $value ) {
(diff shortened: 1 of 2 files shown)


dlh01:
Considering the arithmetic performed on `$value`, as well as the class methods like `get_am_pm()` that are passed `$value` by the default datepicker template and documented to expect integers, I wonder whether it would be preferable to keep at least `$value = (int) $value` rather than removing the whole line.

ravatparmar:
Thanks @dlh01 ,

Added typecasting for `$value`

mslinnea:
@ravatparmar Can you fix the merge conflicts here?  I'd like to include this in the next version release which will be happening soon!

Issue #725, opened 2018-09-13
Clean up PHPCS issues


danielbachhuber:
@kevinfodness It seems like the PHPCS failures could be automatically fixed with `phpcbf`. Want me to submit a pull request for this?

kevinfodness:
Sure!

cc @mboynes

PR #726, opened 2018-09-14
Clean up WordPress coding standards

danielbachhuber:
See #725


danielbachhuber:
@kevinfodness `phpcbf` got us part of the way. Here's what's remaining: https://gist.github.com/danielbachhuber/102edbe73ebbc8484b6e12c856ad4f38

I think we might want to:

1. Fix the minor issues like `json_encode() is discouraged. Use wp_json_encode() instead.`.
2. Ignore `WordPress.PHP.StrictComparisons.LooseComparison` and some other sniffs we know we don't want to change.
3. Give VIP a heads up that we want to land a huge diff of phpcs fixes. It doesn't look like there have been any functional changes since Fieldmanager 1.2.3, so we could keep coding standards changes exclusively to 1.2.4. But, I want to make sure VIP is willing to handle these changes in a timely manner.

bizmate:
hi, just a quick comment, if you are looking to comply with VIP standards it might be great to add 
<rule ref="WordPress-VIP-Go" /> in this PR

danielbachhuber:
@kevinfodness Closing for now. Happy to pick this up again when you're ready to work on it.

PR #731, opened 2018-12-13
Add action hook 'fm_post_before_update_meta'

mae829:
Needed a hook to create/update a separate meta field.

Example:
Event calendar that allows for URL parameter based on `YYY-MM-DD` format to query and display events on that date.

Built two separate `Fieldmanager_Datepicker` fields for an Event, `event_start` and `event_end` saved as UNIX timestamps (default behavior for `Fieldmanager_Datepicker`).

To easily display the date, and query dates through the URL parameter, save/update a new meta field on the fly `event_date` on event date updates.

This hook allows for this, and many other implementations.

### DIFF
--- a/php/context/class-fieldmanager-context-post.php
+++ b/php/context/class-fieldmanager-context-post.php
@@ -310,6 +310,11 @@ class Fieldmanager_Context_Post extends Fieldmanager_Context_Storable {
 	 *                                Default empty.
 	 */
 	protected function update_data( $post_id, $meta_key, $meta_value, $data_prev_value = '' ) {
+		/**
+		 * Last chance to do something with data before update.
+		 */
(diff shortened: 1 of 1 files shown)


dlh01:
Hi @mae829, and thanks for submitting the PR!

What would be the advantage of a using Fieldmanager-specific hook rather than the `added_{$type}_metadata`, `updated_{$type}_metadata`, and `deleted_{$type}_metadata` hooks provided by WordPress core? As a bonus, those hooks work with all meta types out of the box.

mae829:
@dlh01 
I was looking for a way to specifically target Fieldmanager generated meta boxes but I can see why this can be overkill and just a _nice to have_.
Closing request.
Thanks

PR #733, opened 2018-12-31
Fixes an autocomplete search bug

mslinnea:
This fixes an issue where searching for titles that have an apostrophe in them was failing to return results.


mslinnea:
@mboynes Can you review this and see if we can release a new version with this fix and any other priority ones?  Also the failing travis job is unrelated to this change. Looks like something to do with composer, it says: "Your requirements could not be resolved to an installable set of packages."

PR #737, opened 2019-01-11
increment version to 1.2.4

mslinnea:
I will release this version once this is merged.

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -3,7 +3,7 @@
  * Fieldmanager Base Plugin File.
  *
  * @package Fieldmanager
- * @version 1.2.3
+ * @version 1.2.4
  */
 
 /*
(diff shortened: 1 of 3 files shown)


fieldmanager.php
```diff
@@ -3,22 +3,22 @@
...
  */
 
 /*
 Plugin Name: Fieldmanager
 Plugin URI: https://github.com/alleyinteractive/wordpress-fieldmanager
 Description: Add fields to WordPress programatically.
 Author: Alley Interactive
-Version: 1.2.3
+Version: 1.2.4
 Author URI: https://www.alleyinteractive.com/
```
dlh01: Maybe we should update `Author` and `Author URI` while we're here?

mslinnea: Great idea!  I've updated that.

mslinnea:
Thanks @dlh01 Can you please merge?

PR #738, opened 2019-01-25
Add hooks on form element for integrators to hook into.

jenkoian:
By adding these hooks it allows users to do such things as switch to
blog/restore blog, which is particularly useful when using something
like https://github.com/humanmade/network-media-library where all media
will be stored on a single site in the network.


mslinnea:
Hey @jenkoian I've looked over your request here and while I can see the need, this seems better placed in the abstract class class-fieldmanager-field.php, maybe as a companion to the `fm_element_markup_start/end` filters already in `element_markup()`, see: https://github.com/alleyinteractive/wordpress-fieldmanager/blob/0c5fd8439857b84a86781a4867709af56d0f793a/php/class-fieldmanager-field.php#L581

Another option would be to create a subclass in your project that handles the blog switching.

jenkoian:
@mslinnea thanks for the reply! Yeah that makes sense, so do you think an action at the beginning and end of `element_markup()` ?? Could even do it right next to where the filter are applied? Could even go up further and do it in `Fieldmanager_Context::render_field()` something like..

```php
	protected function render_field( $args = array() ) {
		// ...
		do_action( 'fieldmanager_pre_render_field' );
		$field = $this->fm->element_markup( $data );
		do_action( 'fieldmanager_post_render_field' );
		//...
	}
```

WDYT?

mslinnea:
@jenkoian Would it work out better if the hooks were in `Fieldmanager_Context::render_field()`?  The main difference I see is that the pre action would fire before the nonce is created.

jenkoian:
@mslinnea do you want me to update the PR? Does it matter if fired before the nonce creation?

mslinnea:
@jenkoian Hmm... It may cause an issue.  Can you give it a try and see how it works with your use case?

PR #742, opened 2019-03-28
Do you want to delete?

shernandezpl:
It is recommended to add the option to confirm "if you want to delete", the change is added.
Greetings.

### DIFF
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -182,7 +182,9 @@ $( document ).ready( function () {
 	// Handle remove events
 	$( document ).on( 'click', '.fmjs-remove', function( e ) {
 		e.preventDefault();
-		fm_remove( $( this ) );
+		if (confirm('Do you want to delete?')){
+			fm_remove( $( this ) );
+		}
 	} );
 
 	// Handle collapse events


js/fieldmanager.js
```diff
@@ -182,7 +182,9 @@ $( document ).ready( function () {
 	// Handle remove events
 	$( document ).on( 'click', '.fmjs-remove', function( e ) {
 		e.preventDefault();
-		fm_remove( $( this ) );
+		if (confirm('Do you want to delete?')){
```
mboynes: We're not going to move forward with this change, but I'll point out a couple things for the sake of peer code review: this string would need to be internationalized, and the code formatting (whitespace) would need to match the rest of the file.

mboynes:
Hi @shernandezpl, thanks for the PR!

I’m not convinced of this UX as-is. If I’m actively working through a list of items, with a lot of adding and removing, this UX would be disruptive. If I'm deleting a row that hasn't been saved, there's little love lost.

While I don’t care for this particular UX, I agree with the problem you raise. I’ll write up an issue and propose a different UX there and link it to this.

Thanks again!

Issue #679, opened 2017-12-12
Fieldmanager_Select Fields width of 350px is too wide for Sidebar


mslinnea:
Since updating the plugin to version 1.2.0-alpha, I'm seeing an issue with the width of the select fields in the sidebar:

![screen shot 2017-12-12 at 2 19 02 pm](https://user-images.githubusercontent.com/7308162/33911514-76062fc6-df47-11e7-9e20-7cd0400518e0.png)

I've been able to replicate this issue with the following code:

```
function ai_fm_sidebar_test() {
	$fm = new Fieldmanager_Group( array(
		'name' => 'sidebar-select-test',
		'serialize_data' => false,
		'add_to_prefix' => false,
		'children' => array(
			'test-field' => new Fieldmanager_Select( array(
				'first_empty' => true,
				'label' => __( 'Test', 'ai' ),
				'datasource' => new Fieldmanager_Datasource_Term( array(
					'taxonomy' => 'post_tag',
					'append_taxonomy' => false,
					'only_save_to_taxonomy' => true,
				) ),
			) ),
		),
	) );
	$fm->add_meta_box( __( 'Test', 'ai' ), array( 'post' ), 'side', 'low' );
}
add_action( 'fm_post_post', 'ai_fm_sidebar_test' );

```

It looks like the inline styles are specifying the width as 350px, which is wider than the width of the sidebar.

dlh01:
@mslinnea I can replicate this issue only when Chosen is enabled via `type_ahead` (as in your screenshot, it looks like). Do you also see the issue when `'type_ahead' => false`?

mslinnea:
@dlh01 Thanks for checking this out!  The issue only happens when `'type_ahead' => true`.... I accidentally left out that important detail! I'll update the code sample above.

PR #744, opened 2019-04-25
Multi select gutenberg fix

jomurgel:
May address #679, but I was unable to replicate the issue in that task, so I'm unclear.

- Resolves render issue with the chosen select field on Gutenberg-enabled posts. Select fields, in Gutenberg posts, are returning `.is(":visible")` or `:visible` generally, as false on load. Whereas it returns `true` for non-Gutenberg posts. Unclear why that would be, but removing the `:visible` qualifier resolves initializing Chosen.js on the sidebar.
- Removes the width of `350px` and replaces with `max_width: '100%'` which works both in Gutenberg and non-Gutenberg posts to prevent the select dropdown from expanding past its bounds. Chosen already sets the width inline, this prevents miscalculation overages.

### DIFF
--- a/php/class-fieldmanager-select.php
+++ b/php/class-fieldmanager-select.php
@@ -191,9 +191,9 @@ class Fieldmanager_Select extends Fieldmanager_Options {
 				width: '350px'
 			}
 			$('.fm-wrapper').on("fm_added_element fm_collapsible_toggle fm_activate_tab",".fm-item",function(){
-				$(".chosen-select:visible",this).chosen(chosenOpts)
+				$(".chosen-select",this).chosen(chosenOpts)
 			});
-			$(".chosen-select:visible").chosen(chosenOpts);
+			$(".chosen-select").chosen(chosenOpts);
 		});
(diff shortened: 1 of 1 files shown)


GFargo:
Hey @jomurgel,

Looks like this issue also extends to single autocomplete fields as well.  By removing `:visible` from `fieldmanager-automcomplete.js` [here](https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/js/fieldmanager-autocomplete.js#L12) resolves the issue there as well.

jomurgel:
PR #729 might be a better solution for resolving the issues with the autocomplete field/sortable field behavior on load. A replacement for that part of this PR vs removing the `:visible` selector check right off.

There was quite a bit of discussion about the implications around removing that check altogether, though in my testing, that did not seem to have any real-world effects on rendering/initting fields in Gutenberg or 4.9- WP installs with fields shown or hidden.

kevinfodness:
Fixed via #757

PR #745, opened 2019-04-29
Gutenberg Shim

GFargo:
- Modifies `fm_add_script` JavaScript dependencies to include `wp-edit-post` when Gutenberg is active.
- Includes new script to manually trigger FM `fm_added_element` event via `wp.domReady`

### DIFF
--- a/js/fieldmanager-gutenberg-support.js
+++ b/js/fieldmanager-gutenberg-support.js
@@ -1,0 +1,8 @@
+( function( $ ) {
+	if (wp.domReady) {
+		wp.domReady(function() {
+			console.log('Triggering FM Collapse Toggle Event for Gutenberg.');
+			$(document).trigger('fm_added_element');
+		});
+	}
+} ) ( jQuery );
--- a/js/fieldmanager-gutenberg-support.js
+++ b/js/fieldmanager-gutenberg-support.js
@@ -1,8 +1,10 @@
 ( function( $ ) {
 	if (wp.domReady) {
 		wp.domReady(function() {
-			console.log('Triggering FM Collapse Toggle Event for Gutenberg.');
 			$(document).trigger('fm_added_element');
+
+			// Multiselects listen to fm-wrapper for the event.
+			$('.fm-wrapper').trigger('fm_added_element');
 		});
 	}
 } ) ( jQuery );
--- a/php/util/class-fieldmanager-util-gutenberg.php
+++ b/php/util/class-fieldmanager-util-gutenberg.php
@@ -23,10 +23,23 @@ class Fieldmanager_Util_Gutenberg {
 	 * @return array
 	 */
 	public function add_gutenberg_js_deps( $scripts ) {
-		$screen = get_current_screen();
+		$is_gutenberg_editor = false;
+
+		 // Do we have access to current screen?
+		if ( did_action( 'current_screen' ) ) {
+			$current_screen = get_current_screen();
+			$is_gutenberg_editor = $current_screen instanceof WP_Screen ? $current_screen->is_block_editor : false;
+		}
+
+		// Fallback if we don't have access to `current_screen`.
+		if ( ! $is_gutenberg_editor ) {
+			// Go into globals for post ID on new auto-draft posts.
+			$post_id = get_the_ID() ? get_the_ID() : $GLOBALS['post_ID'];
+			$is_gutenberg_editor = use_block_editor_for_post( $post_id );
+		}
 
 		// If we are working within the context of the block editor, we should ensure required deps are loaded.
-		if ( $screen->is_block_editor ) {
+		if ( $is_gutenberg_editor ) {
 			foreach ( $scripts as $index => $script ) {
 				$scripts[ $index ]['deps'][] = 'wp-edit-post';
 			}
--- a/php/util/class-fieldmanager-util-gutenberg.php
+++ b/php/util/class-fieldmanager-util-gutenberg.php
@@ -39,7 +39,15 @@ class Fieldmanager_Util_Gutenberg {
 
 		// Fallback if we don't have access to `current_screen`.
 		if ( ! $is_gutenberg_editor ) {
-			$post_id = false !== get_the_ID() ? get_the_ID() : $GLOBALS['post_ID'];
+			if ( false !== get_the_ID() ) {
+				$post_id = get_the_ID();
+			} elseif ( isset( $GLOBALS['post_ID'] ) ) {
+			        // Sanatize superglobal.
+				$_GET['post_ID'] = filter_input(INPUT_GET, 'post_ID', FILTER_SANITIZE_STRING);
+				$post_id = $GLOBALS['post_ID'];
+			} else {
+				$post_id = null;
+			}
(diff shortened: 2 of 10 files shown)


js/fieldmanager-gutenberg-support.js
```diff
@@ -0,0 +1,8 @@
+( function( $ ) {
+	if (wp.domReady) {
+		wp.domReady(function() {
+			console.log('Triggering FM Collapse Toggle Event for Gutenberg.');
```
GFargo: _**Note:**  For testing purposes, will be removed before merge._

php/util/class-fieldmanager-util-gutenberg.php
```diff
@@ -0,0 +1,52 @@
...
+		 // Do we have access to current screen?
+		if ( did_action( 'current_screen' ) ) {
+			$current_screen = get_current_screen();
+			$is_gutenberg_editor = $current_screen instanceof WP_Screen ? $current_screen->is_block_editor : false;
+		}
+
+		// Fallback if we don't have access to `current_screen`.
+		if ( ! $is_gutenberg_editor ) {
+			// Go into globals for post ID on new auto-draft posts.
+			$post_id = get_the_ID() ? get_the_ID() : $GLOBALS['post_ID'];
```
GFargo: Is there a preferred way to get at the post ID on _auto-draft_ posts?

php/util/class-fieldmanager-util-gutenberg.php
```diff
@@ -0,0 +1,52 @@
...
+		$is_gutenberg_editor = false;
+
+		 // Do we have access to current screen?
+		if ( did_action( 'current_screen' ) ) {
+			$current_screen = get_current_screen();
+			$is_gutenberg_editor = $current_screen instanceof WP_Screen ? $current_screen->is_block_editor : false;
+		}
+
+		// Fallback if we don't have access to `current_screen`.
+		if ( ! $is_gutenberg_editor ) {
```
dlh01: This block will still run even when the screen is available if `$current_screen->is_block_editor` is `false`.

php/util/class-fieldmanager-util-gutenberg.php
```diff
@@ -0,0 +1,52 @@
...
+			$post_id = false !== get_the_ID() ? get_the_ID() : $GLOBALS['post_ID'];
```
dlh01: - The use of a template tag seems unexpected here. What is the desired value? The ID of the post being edited?
- I think `$GLOBALS['post_ID']` will exist only only in `$_POST`. For `$_GET` requests, the ID is in `$_GET['post']`. So, there is a possibility of undefined-index warnings, and I'm not sure whether the logic here captures all the needed cases.
- Also, any value from a superglobal will need sanitization.
- Having said all that: I think `fm_enqueue_scripts` will also run on the frontend if any Fieldmanager assets are enqueued (via `\Fieldmanager_Util_Assets::hook_enqueue()`). If so, that could lead to cases where `current_screen` does not fire, but because the global post from `get_the_ID()` supports the block editor, `wp-post-edit` will be enqueued. That might be the intention here — maybe it's why `get_the_ID()` is used?

php/util/class-fieldmanager-util-gutenberg.php
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Gutenberg helpers.
+ */
+class Fieldmanager_Util_Gutenberg {
+	public function __construct() {
+		add_filter( 'fm_enqueue_scripts', [ $this, 'add_gutenberg_js_deps' ], 99 );
+
+		// Gutenberg sidebar polyfill.
+		fm_add_script( 'fieldmanager-gutenberg-polyfill', 'js/fieldmanager-gutenberg-support.js' );
```
dlh01: My memory is fuzzy here: Will `fm_add_script()` cause this script to be enqueued? If so, is it now being enqueued on all requests by nature of its being in the constructor?

php/util/class-fieldmanager-util-gutenberg.php
dlh01: > but because the global post from get_the_ID() supports the block editor, wp-post-edit will be enqueued

Pardon me: Not enqueued, but added as a dependency. However, if a script dependency is not enqueued, is the dependency enqueued, or does the script fail to load?

GFargo: We need to refactor how/when we're accessing `$GLOBALS['post_ID']`.  This creates warnings on the WP admin dashboard.

jomurgel: @GFargo could probably do something like this to check to make sure that `$GLOBALS` is a thing.

```suggestion
			$post_id = false !== get_the_ID() ? get_the_ID() : isset( $GLOBALS['post_ID'] ) ? $GLOBALS['post_ID'] : null;
```

php/util/class-fieldmanager-util-gutenberg.php
jomurgel: @dlh01 This function triggers `pre_enqueue_script` which checks for `did_action( 'admin_enqueue_scripts' ) || did_action( 'wp_enqueue_scripts' )` which enqueues the script else adds the script handle the to the scripts array and runs `$this->hook_enqueue()` which adds actions for `admin_enqueue_scripts` and `wp_enqueue_scripts` and filters/enqueues there.

php/util/class-fieldmanager-util-gutenberg.php
jomurgel: Even if `$current_screen->is_block_editor` is false, we'd hit the fallback which returns false also in my testing, which prevented anything from loading. 🤔 This seems like what it's supposed to do, no?

dlh01: Sure, I just mean that running the fallback seems redundant if `WP_Screen` already provided the answer.

php/util/class-fieldmanager-util-gutenberg.php
dlh01: Right. So, it looks like the script is loading on all requests regardless of whether a field is on the page. It would automatically load on the frontend as well, but there's an `is_admin()` check in `\Fieldmanager_Util_Assets::add_script()` to prevent that from happening.

php/util/class-fieldmanager-util-gutenberg.php
jomurgel: Oh, right. I agree. Tbh I'm not certain why we have the fallback specifically since all posts should have access to it, but @GFargo might have more information for that.

jomurgel:
Made several updates here to tackle the following:
- Add a fallback for an older version of WP to avoid errors trying to access functions/properties that don't exist.
- Resolves PHPCS and PHPUnit errors.
- Addressed several notes above.

js/fieldmanager-gutenberg-support.js
```diff
@@ -0,0 +1,10 @@
+( function( $ ) {
+	if (wp.domReady) {
```
jameswburke: `wp` isn't always available when this fires.

kevinfodness:
Fixed via #757

PR #749, opened 2019-06-03
v. 1.2.5

shawncook:
Drafts a release version 1.2.5. Includes bug fixes and test changes.

- Fix checkboxes validating with display_if (498-test-against-all-checkboxes)
- Fix autocomplete validation (autocomplete-validation)
- Bump `grunt-contrib-qunit` version to fix a failing test
- Remove a test: "The 5.3 test running against `latest` should be removed because WP 5.2 dropped support for PHP versions lower than 5.6."


php/util/class-fieldmanager-util-validation.php
```diff
@@ -222,9 +222,11 @@ public function add_validation() {
 			}
 
 			// Fields that should always be ignored.
-			$ignore[] = '.fm-autocomplete';
+			$ignore = array();
```
jomurgel: This could be just `[]` to follow the PHP 7+ conventions.

jomurgel:
Might have to restart the build to make Travis report. 😐

php/util/class-fieldmanager-util-validation.php
```diff
@@ -356,8 +357,8 @@ private function format_value( $value ) {
 	 * @return string The field name with quotes added if necessary
 	 */
 	private function quote_field_name( $field ) {
-		// Check if the field name is alphanumeric (underscores and dashes are allowed).
-		if ( ctype_alnum( str_replace( array( '_', '-' ), '', $field ) ) ) {
+		// Check if the field name is alphanumeric (underscores are allowed).
+		if ( ctype_alnum( str_replace( array( '_' ), '', $field ) ) ) {
```
mboynes: Since the search is no longer multiple strings, it should just be a string vs an array

php/util/class-fieldmanager-util-validation.php
mboynes: FM supports PHP <7, so we should not use PHP7 syntax

php/util/class-fieldmanager-util-validation.php
```diff
@@ -222,9 +221,11 @@ public function add_validation() {
 			}
 
 			// Fields that should always be ignored.
-			$ignore[] = '.fm-autocomplete';
+			$ignore = array();
 			$ignore[] = "input[type='button']";
-			$ignore[] = ':hidden';
+
+			// Fields that contain hidden inputs and still need to be verified, ie: image fields.
+			$ignore[] = ':not(\'.fm-media-button, .fm-autocomplete\') + :hidden';
```
mboynes: * [ ] Nitpicky, but this string should be wrapped in double-quotes so that the single quotes don't have to be escaped
* [ ] I'm not fully grocking the ramifications of this selector. Would a hidden field that _doesn't_ have a sibling still get selected? Does this require that a hidden field have a sibling?

shawncook: @mboynes This change, instead of targeting & ignoring all hidden fields, is meant to only target hidden items not preceded by an "add media" button or an autocomplete field. In other words, it's still targeting all hidden fields, but excludes ones with a previous sibling of autocomplete or media button. 

One of the bugs I was trying to fix in this PR was to ensure that these two field types were validating correctly -- targeting the field itself wasn't helpful, because the value of the field is written to a hidden sibling that's used when the form is submitted. Before this change, validation would fail silently for those two types.

shawncook: >Would a hidden field that doesn't have a sibling still get selected? Does this require that a hidden field have a sibling?

Going to test this again quick, but I think it'll get all hidden fields that are not preceded by elements with those classes, and hidden fields that stand alone would be included because they're not preceded by those classes.

shawncook: You were right -- that was excluding standalone :hidden fields. I added :hidden:first-child, which should target those.

PR #750, opened 2019-06-21
Adds Button Field Type

jomurgel:
Helps address: #70 
- Adds button field support for Field Manager.

Use:
```
  "FIELDNAME": {
    "label": { "__": "Button Label" },
    "type": "button",
    "field_class": "CUSTOM_CLASSNAME",
    "options": {
      "button_text": { "__": "Button Text" },
    }
  },
```

### DIFF
--- a/php/class-fieldmanager-button.php
+++ b/php/class-fieldmanager-button.php
@@ -12,6 +12,9 @@ class Fieldmanager_Button extends Fieldmanager_Options {
 
 	/**
 	 * Setup Button Template and Type.
+	 *
+	 * @param string $label   Field label.
+	 * @param array  $options The form options.
 	 */
 	public function __construct( $label = '', $options = array() ) {
 		$this->input_type = 'button';
--- a/php/class-fieldmanager-button.php
+++ b/php/class-fieldmanager-button.php
@@ -9,6 +9,12 @@
  * Button Field
  */
 class Fieldmanager_Button extends Fieldmanager_Options {
+	/**
+	 * Button content string.
+	 *
+	 * @var string
+	 */
+	public $button_content = '';
 
 	/**
 	 * Setup Button Template and Type.
@@ -17,8 +23,8 @@ class Fieldmanager_Button extends Fieldmanager_Options {
 	 * @param array  $options The form options.
 	 */
 	public function __construct( $label = '', $options = array() ) {
-		$this->input_type = 'button';
-		$this->template = fieldmanager_get_template( 'button' );
+		$this->input_type     = 'button';
+		$this->template       = fieldmanager_get_template( 'button' );
 		parent::__construct( $label, $options );
 	}
 }
--- a/php/class-fieldmanager-button.php
+++ b/php/class-fieldmanager-button.php
@@ -6,7 +6,7 @@
  */
 
 /**
- * Button Field
+ * Button Field.
  */
 class Fieldmanager_Button extends Fieldmanager_Options {
 	/**
@@ -20,7 +20,7 @@ class Fieldmanager_Button extends Fieldmanager_Options {
 	 * Setup Button Template and Type.
 	 *
 	 * @param string $label   Field label.
-	 * @param array  $options The form options.
+	 * @param array  $options The field options.
 	 */
 	public function __construct( $label = '', $options = array() ) {
 		$this->input_type     = 'button';
--- a/php/class-fieldmanager-button.php
+++ b/php/class-fieldmanager-button.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * Class file for Fieldmanager_Button
+ *
+ * @package Fieldmanager
+ */
+
+/**
+ * Button Field
+ */
+class Fieldmanager_Button extends Fieldmanager_Options {
+
+	/**
+	 * Setup Button Template and Type.
+	 */
+	public function __construct( $label = '', $options = array() ) {
+		$this->input_type = 'button';
+		$this->template = fieldmanager_get_template( 'button' );
+		parent::__construct( $label, $options );
+	}
+}
--- a/php/class-fieldmanager-field.php
+++ b/php/class-fieldmanager-field.php
@@ -96,6 +96,20 @@ abstract class Fieldmanager_Field {
 	public $label_after_element = false;
 
 	/**
+	 * Button content for the form element.
+	 *
+	 * @var string
+	 */
+	public $button_content = '';
+
+	/**
+	 * If true, the button content will be displayed after the element.
+	 *
+	 * @var bool
+	 */
+	public $button_content_after_element = true;
+
+	/**
 	 * Description for the form element.
 	 *
 	 * @var string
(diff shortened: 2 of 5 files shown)


templates/button.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @package Fieldmanager
+ */
+?>
+<input
+	class="fm-button-element button-secondary <?php echo esc_attr( $this->field_class ); ?>"
+	type="button"
+	name="<?php echo esc_attr( $this->get_form_name() ); ?>"
+	id="<?php echo esc_attr( $this->get_element_id() ); ?>"
+	value="<?php echo esc_attr( $this->options['button_text'] ); ?>"
```
mboynes: This isn't the way `options` is used, that field is for e.g. `<select>` options. Instead, `Fieldmanager_Button` should get a public property, `$button_text`, and then it would be set like any other field argument

templates/button.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Default template for Fieldmanager_Button
+ *
+ * @package Fieldmanager
+ */
+?>
+<input
+	class="fm-button-element button-secondary <?php echo esc_attr( $this->field_class ); ?>"
```
mboynes: The button type should be an argument rather than hardcoded `button-secondary`. That can be the default, but making it an argument gives the implementor the choice.

templates/button.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Default template for Fieldmanager_Button
+ *
+ * @package Fieldmanager
+ */
+?>
+<input
```
mboynes: Something to consider: instead of `input[type=button]`, this could be a `button[type=button]`. If we did that, it would mean that the `button_text` could instead be `button_content` and could theoretically be any html content, like an image (we'd have to piggyback on FM's escaping argument to escape html by default, but allow for an override). I could be convinced either way; I'm curious what others think about it.

templates/button.php
jomurgel: Ah, that makes sense. 👍

templates/button.php
mboynes: I refreshed my own memory how it works, and the escaping if we opted for `button` would be really simple:

```php
<button
    type="button"
    ...
>
    <?php echo $this->escape( 'button_content' ); ?>
</button>
```

If an implementor wanted to allow rich text in the field, they'd simply pass something like `[ 'escape' => [ 'button_content' => 'wp_kses_post' ] ]` in their field definition.

jomurgel:
Made several updates based on the notes above including converting the input to a button and taking advantage of the `$this->escape()` function which would provide the ability to include HTML inside the button. Tested in practice, and added a unit test for that. I think this makes more sense, but definitely need another set of eyes on the code since I'm less familiar with FieldManager.

``` json
  "FIELDNAME": {
    "label": { "__": "LABEL TEXT" },
    "type": "button",
    "button_content": { "__": "BUTTON CONTENT" },
    "attributes": {
      "class": "additonal-class-names",
    },
    "escape": {
      "button_content": "wp_kses_post"
    }
  }
```

tests/php/test_fieldmanager_field.php
```diff
@@ -925,6 +925,29 @@ public function test_description_escaping() {
 		$this->assertContains( $description_html, $html );
 	}
 
+	public function test_button_content_escaping() {
+		$id                  = rand_str();
+		$button_content_raw  = rand_str();
```
mboynes: I'm assuming this was cribbed from another test, but we've since stopped using `rand_str()`. It's better to use static strings, since unit tests should always be as reliable and repeatable as possible.

tests/php/test_fieldmanager_field.php
```diff
@@ -925,6 +925,29 @@ public function test_description_escaping() {
...
+		$button_content_html = "<strong id='{$id}'>{$button_content_raw}</strong>";
+		$args                = array(
+			'name'        => 'button_content_escape_testing',
+			'description' => $button_content_html,
+		);
+
+		// Ensure that, by default, the description is present without the HTML
+		$field = new Fieldmanager_TextField( $args );
```
mboynes: This should be `Fieldmanager_Button`, right? (same below)

tests/php/test_fieldmanager_field.php
```diff
@@ -925,6 +925,29 @@ public function test_description_escaping() {
 		$this->assertContains( $description_html, $html );
 	}
 
+	public function test_button_content_escaping() {
+		$id                  = rand_str();
+		$button_content_raw  = rand_str();
+		$button_content_html = "<strong id='{$id}'>{$button_content_raw}</strong>";
+		$args                = array(
+			'name'        => 'button_content_escape_testing',
+			'description' => $button_content_html,
```
mboynes: This should be `button_content` instead of `description`

php/class-fieldmanager-field.php
```diff
@@ -95,6 +95,20 @@ abstract class Fieldmanager_Field {
 	 */
 	public $label_after_element = false;
 
+	/**
+	 * Button content for the form element.
+	 *
+	 * @var string
+	 */
+	public $button_content = '';
```
mboynes: This should go into `class-fieldmanager-button.php`. Adding it here means that any field can set `button_text`, which isn't the case.

php/class-fieldmanager-button.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Button Field
+ */
+class Fieldmanager_Button extends Fieldmanager_Options {
+
+	/**
+	 * Setup Button Template and Type.
+	 *
+	 * @param string $label   Field label.
+	 * @param array  $options The form options.
```
mboynes:
```suggestion
	 * @param array  $options The field options.
```

php/class-fieldmanager-button.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * Class file for Fieldmanager_Button
+ *
+ * @package Fieldmanager
+ */
+
+/**
+ * Button Field
```
mboynes:
```suggestion
 * Button Field.
```

This should end in a full stop

php/class-fieldmanager-field.php
```diff
@@ -694,6 +708,10 @@ public function single_element_markup( $value = null, $is_proto = false ) {
 			}
 		}
 
+		if ( isset( $this->button_content ) && ! empty( $this->button_content ) && ! $this->button_content_after_element ) {
+			$out .= sprintf( '<div class="fm-item-button_content">%s</div>', $this->escape( 'button_content' ) );
+		}
+
```
mboynes: As discussed elsewhere, this should be removed

php/class-fieldmanager-field.php
```diff
@@ -95,6 +95,20 @@ abstract class Fieldmanager_Field {
...
+	 */
+	public $button_content = '';
+
+	/**
+	 * If true, the button content will be displayed after the element.
+	 *
+	 * @var bool
+	 */
+	public $button_content_after_element = true;
+
```
mboynes: As discussed elsewhere, this should be removed

jomurgel:
Closing due to inactivity.

dlh01:
I'm sorry that this one hasn't gotten over the line yet, @jomurgel. Mind if we keep the PR and branch open, though, so that someone else can pick it up? Based on Matt's last review, there isn't much more left to do.

PR #761, opened 2019-11-12
Add a test for local times saved across DST change

### LINKED ISSUE #759: Datepicker, store_local_time and daylight saving
We had an issue pop up recently when editing date/time in a custom field when the date and time happened after the 1hr change on daylight saving. The WordPress install is set to New York time, `store_local_time` is enabled. To replicate this, change the date/time of the event to anything after March 8, 2020 (the next time change), the hour gets reduced by one each time the post is saved (entered 2pm, saved it and it's 1pm, save it again and it's noon).

I was also able to replicate it when setting the date/time to prior to the recent 1hr time change (Nov 3rd). This seems to only happen when editing a time that falls outside the six-month window when there is no time change.

Some sample to drop into a dev site in order to replicate the issue.
```
<?php
/**
 * Plugin Name: Test Fieldmanager Time
 */

add_action( 'fm_post_post', 'test_fieldmanager_date_time' );

function test_fieldmanager_date_time() {

	$fields = [];

	$fields['start-time'] = new \Fieldmanager_Datepicker(
		[
			'label'            => __( 'Start Date/Time (Eastern)', 'stadium' ),
			'use_time'         => true,
			'store_local_time' => true,
			'default_value'    => current_time( 'timestamp', true ),
		]
	);

	$fm = new \Fieldmanager_Group(
		[
			'name'           => 'event-settings',
			'serialize_data' => false,
			'children'       => $fields,
		]
	);

	$fm->add_meta_box( 'Live Event Settings', [ 'post' ] );
}
```

mboynes:
Resolves #759

Makes `Fieldmanager_Datepicker::form_element()` apply a DST-aware timezone offset when rendering a `store_local_time` value, so it mirrors `presave()`'s `get_gmt_from_date()` and a no-change re-save no longer drifts the stored time by an hour across a DST boundary.

Also updates the label/description escaping and rich-text save tests to expect double-quoted HTML attribute delimiters, matching how current WordPress core's `wp_kses` now normalizes attribute output.

### DIFF
--- a/php/class-fieldmanager-datepicker.php
+++ b/php/class-fieldmanager-datepicker.php
@@ -84,6 +84,38 @@ class Fieldmanager_Datepicker extends Fieldmanager_Field {
 	}
 
 	/**
+	 * Get the current site's local timezone as a DateTimeZone.
+	 *
+	 * This is a wrapper/shim for `wp_timezone()`, introduced in WordPress 5.3.
+	 *
+	 * @return \DateTimeZone
+	 */
+	protected function get_local_datetimezone() {
+		if ( function_exists( 'wp_timezone' ) ) {
+			return wp_timezone();
+		}
+
+		/* If wp_timezone() isn't available, shim it. */
+
+		$timezone_string = get_option( 'timezone_string' );
+
+		if ( $timezone_string ) {
+			return new \DateTimeZone( $timezone_string );
+		}
+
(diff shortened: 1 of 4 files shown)


kevinfodness (approved):
I checked the implementation of the shim against the current code for the WP core timezone functions and it still matches. Hopefully no one is still running WP <= 5.2.x but good to know this library will still work if they are :) 🍣

PR #773, opened 2020-06-16
Add Menu Item Context

kjbenk:
Adds a new context for menu items since a hook as been added into core to make this possible - https://github.com/WordPress/WordPress/blob/26bda18a23174afb048afbe62296c76a62add542/wp-admin/includes/class-walker-nav-menu-edit.php#L242

### DIFF
--- a/tests/php/test-fieldmanager-context-menuitem.php
+++ b/tests/php/test-fieldmanager-context-menuitem.php
@@ -11,18 +11,10 @@ class Test_Fieldmanager_Context_Menuitem extends WP_UnitTestCase {
 		parent::setUp();
 		Fieldmanager_Field::$debug = true;
 
-		$this->post = array(
+		$this->post = self::factory()->post->create_and_get( array(
 			'post_type'    => 'nav_menu_item',
 			'post_status'  => 'publish',
-			'post_content' => rand_str(),
-			'post_title'   => rand_str(),
-		);
-
-		// insert a post
-		$this->post_id = wp_insert_post( $this->post );
-
-		// reload as proper object
-		$this->post = get_post( $this->post_id );
+		) );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


tests/php/test-fieldmanager-context-menuitem.php
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @group context
+ * @group menuitem
+ */
+class Test_Fieldmanager_Context_Menuitem extends WP_UnitTestCase {
+	public function setUp() {
+		parent::setUp();
+		Fieldmanager_Field::$debug = true;
+
+		$this->post = array(
```
kevinfodness, line 14: This would be simplified greatly by doing `self::factory()->post->create_and_get(` which obviates the need for the `get_post` below, and handles the post insertion. It will also randomize the post title and content for you, so you can eliminate that. In a few places below, you're already using `$this->post->ID` so you can just do that instead of `$this->post_id`.

tests/php/test-fieldmanager-context-menuitem.php
```diff
@@ -0,0 +1,176 @@
...
+
+		// Only run these tests for WP versions above 5.4.0.
+		if ( version_compare( $wp_version, '5.4.0', '<' ) ) {
+			return;
+		}
+
+		$base = $this->_get_elements();
+		$base->add_nav_menu_fields();
+
+		$post_id = wp_insert_post(
```
kevinfodness, line 161: Same comment as above.

tests/php/test-fieldmanager-context-menuitem.php
```diff
@@ -0,0 +1,176 @@
...
+		$this->assertEquals( count( $saved_value['test_extended'] ), 4 );
+		$this->assertEquals( count( $saved_value['test_extended'][0]['extext'] ), 1 );
+		$this->assertEquals( count( $saved_value['test_extended'][1]['extext'] ), 3 );
+		$this->assertEquals( count( $saved_value['test_extended'][2]['extext'] ), 1 );
+		$this->assertEquals( count( $saved_value['test_extended'][3]['extext'] ), 1 );
+		$this->assertEquals( $saved_value['test_extended'][1]['extext'], array( 'second1', 'second2', 'second3' ) );
+		$this->assertEquals( $saved_value['test_extended'][3]['extext'][0], 'fourth' );
+	}
+
+	public function test_programmatic_save_posts() {
```
kevinfodness, line 150: This test doesn't contain any assertions.

kjbenk: Good catch :) I was just following this code, but I did not vet it out entirely - https://github.com/alleyinteractive/wordpress-fieldmanager/blob/1a402804ba5d89d524a28b44fbd1e3f13463d63a/tests/php/test-fieldmanager-context-post.php#L135

PR #781, opened 2020-10-20
Implement Fieldmanager_Content, Plaintext, HTML, and Markdown

jameswburke:
## Summary
Implements an abstract `Fieldmanager_Content` class, and 3 implementations that supports rendering static text (`Fieldmanager_Content_Plaintext`), html (`Fieldmanager_Content_Markup`), and markdown (`Fieldmanager_Content_Markdown`).

## Notes for reviewers
Questions for reviewer,
* Should `Parsedown()` be included differently? Perhaps composer, or a different directory?
* Any tips for how to better structure or use existing Field functionality?
* Any thoughts on additional tests to implement?

## Changelog entries
* **Added**: `Fieldmanager_Content` abstract class.
* **Added**: `Fieldmanager_Content_Plaintext`, `Fieldmanager_Content_Markup`, and `Fieldmanager_Content_Markdown` classes and unit tests.

## Ticket
https://alleyinteractive.atlassian.net/browse/FLAME-113


php/class-fieldmanager-content.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @var string
+	 */
+	public $content = '';
+
+	/**
+	 * Type of content being rendered. plaintext, html, markdown supported.
+	 *
+	 * @var string
+	 */
+	public $content_type = 'plaintext';
```
mboynes, line 32: Proposal for consideration/discussion: instead of `$content_type` and `$content`, have `$plaintext`, `$markup`, and `$markdown`. Examples:

```php
new Fieldmanager_Content( [ 'markdown' => 'Some _Markdown_' ] );
new Fieldmanager_Content( [ 'markup' => 'Some <em>Markup</em>' ] );
```

fieldmanager.php
```diff
@@ -98,6 +98,7 @@ function fieldmanager_load_file( $file ) {
 }
 
 // Load utility classes with helper functions. These can't be autoloaded.
+fieldmanager_load_file( 'libraries/parsedown/Parsedown.php' );
```
mboynes, line 101: If a site is already using Parsedown, this will throw an exception and potentially take the site down. We'll need to namespace it. Further, it should be able to be autoloaded since it's a class.

mboynes (changes requested):

fieldmanager.php
dlh01: If it's autoloaded, does it also need to be namespaced? (I get that they're not mutually exclusive on a technical level; more wondering whether autoloading is sufficient due diligence.)

mboynes: good question. two whatifs come to mind:

1. A page could theoretically be loading Parsedown very late in the request, e.g. for something in the admin footer. if this loaded first using the autoloader, it could still break that integration. I think.
2. If a site already uses Parsedown, it might be using a different version. if ours loaded first, that could produce unexpected results in the other integration; if theirs loaded first, it could produce unexpected results in Fieldmanager.

If my hunches are right, I suspect it would not be sufficient due diligence. But I'm very prepared to be wrong 😉 .

dlh01: > A page could theoretically be loading Parsedown very late in the request, e.g. for something in the admin footer. if this loaded first using the autoloader, it could still break that integration. I think.

True, if Parsedown was being loaded directly via `require` or `include` without first checking whether the class was defined. In that case, though, the developer would be at risk of the same issue from any number of libraries that might include Parsedown as a dependency.

> If a site already uses Parsedown, it might be using a different version. if ours loaded first, that could produce unexpected results in the other integration; if theirs loaded first, it could produce unexpected results in Fieldmanager.

Sold!

dlh01:
Some drive-by comments...

- The name `Fieldmanager_Content` potentially implies some sort of relationship to `post_content` that might prove confusing. I don't have a great alternative to offer, though. `Fieldmanager_Display`? `Fieldmanager_InBetween`?

- I think it would be great if the field could accept the path to a file as an alternative to accepting the content string, as it would allow a developer to maintain a separate `.md` file rather than have to manage the content within PHP variables.

- Along the lines of Matt's suggestion: There's currently one class responsible for handling three content types, but I wonder whether the developer experience would be improved by instead having three classes, one for each content type. For maintainers, these classes would likely be smaller and thus have less to maintain. For developers, only one thing needs to be specified to use the field (the class) instead of two (the class and the type property). Lastly, FM already has abstract classes for similar situations (e.g. `Fieldmanager_Options`).

php/class-fieldmanager-content.php
jameswburke: I like this, but ended up going with @dlh01's suggestion to break out into multiple fields (I had debated that approach to begin with), `Fieldmanager_Plaintext`, `Fieldmanager_Markup`, and `Fieldmanager_Markdown`. As a result, each one uses `content` still to maintain some consistency.

```php
new Fieldmanager_Plaintext( [ 'content' => 'Some Text' ] );
new Fieldmanager_Markup( [ 'content' => 'Some <em>Markup</em>' ] );
new Fieldmanager_Markdown( [ 'content' => 'Some _Markdown_' ] );
```

fieldmanager.php
jameswburke: I've implemented autoloading and namespaced to `Fieldmanager_Parsedown`.

The conditional is a bit clunky (see below), but I'm open to suggestions on how it should be structured. I added an explanation/call for refactoring should someone need a broader solution/autoload pattern.

Normally I'd namespace and autoload based on the path (`Fieldmanager/Libraries/Parsedown` would look in `/php/libraries/parsedown/class-parsedown.php`), but it's not a pattern anywhere else in FM, and I wanted to stay consistent.

```
/**
 * Autoload support for Parsedown.
 *
 * This is done specifically for Parsedown because it's a third party
 * library, which presents a different folder structure than other
 * autoloads. If other libraries are introduced in the future, please
 * abstract/refecator/generalize this functionality as needed.
 */
if ( 'Fieldmanager_Parsedown' === $class ) {
	return fieldmanager_load_file( 'libraries/parsedown/class-fieldmanager-parsedown.php' );
}
```

jameswburke:
@dlh01, thanks David, I've split out `Fieldmanager_Content` into `Fieldmanager_Plaintext`, `Fieldmanager_Markup`, and `Fieldmanager_Markdown`. I had originally gone down this path, but found it was possibly too verbose, but your suggestion encouraged me to revisit. I do think it's a better developer experience at the end of the day, even at the expense of some additional code to maintain.

I like your point about loading files, but I think `file_get_contents` is a simple way to achieve that. I wouldn't add support for anything more complex as part of this initial implementation.

```php
new Fieldmanager_Markdown(
	[
		'content' => file_get_contents( __DIR__ . '/readme.md' ),
	]
);
```

fieldmanager.php
```diff
@@ -74,6 +74,18 @@ function fieldmanager_load_class( $class ) {
...
+	 * Autoload support for Parsedown.
+	 *
+	 * This is done specifically for Parsedown because it's a third party
+	 * library, which presents a different folder structure than other
+	 * autoloads. If other libraries are introduced in the future, please
+	 * abstract/refecator/generalize this functionality as needed.
+	 */
+	if ( 'Fieldmanager_Parsedown' === $class ) {
+		return fieldmanager_load_file( 'libraries/parsedown/class-fieldmanager-parsedown.php' );
+	}
```
mboynes, line 87: I would favor the FQN being `\Fieldmanager\Libraries\Parsedown\Parsedown` and this condition could then be updated to check if the namespace starts with `Fieldmanager\Libraries\`, and if so, load the file in `libraries/**/class-*.php` based on the rest of the namespace and the class name. this would support us adding other libraries in the future. No time like the present to adopt newer standards!

dlh01:
`file_get_contents()` certainly works, and it isn't mutually exclusive with adding a `file` property later. The downside to consider is performance: The file has to be read from disk immediately rather than only in the moment that Fieldmanager needs it.

jameswburke:
Updated autoload approach to support a `Fieldmanager\Library\Package\Class` format, and implemented `abstract class Fieldmanager_Content` and refactored for other content fields.

jameswburke:
@mboynes & @dlh01, could you take another look at my latest pass on this? Thanks!

fieldmanager.php
```diff
@@ -74,6 +74,12 @@ function fieldmanager_load_class( $class ) {
 		return fieldmanager_load_file( 'util/class-fieldmanager-util-' . $class_id . '.php' );
 	}
 
+	if ( 0 === strpos( $class, 'Fieldmanager\Libraries' ) ) {
+		$classes  = explode( '\\', $class );
+		$class_id = strtolower( array_pop( $classes ) );
```
mboynes, line 79: while not critical for this PR, it would behoove us to be forward-thinking in this change and support n-depth namespaces/classes, so if the next library had the class `Fieldmanager\Libraries\Grand_Parent\Parent\Child` it would automatically be loaded at `libraries/grand-parent/parent/class-child.php`. We have lots of art for this, e.g. (parts of) the autoload generator in [mantle](https://github.com/alleyinteractive/mantle-framework/blob/main/src/autoload.php).

php/class-fieldmanager-content.php
```diff
@@ -0,0 +1,51 @@
...
+		return $this->render_content( (string) $this->content );
+	}
+
+	/**
+	 * Render content.
+	 *
+	 * @param string $content Content to render.
+	 * @return string Rendered content.
+	 */
+	abstract protected function render_content( string $content ): string;
```
mboynes, line 50: * The implementing classes need to match this signature, so it needs to be public.
* Even though Fieldmanager says it requires PHP 7, we also explicitly say "Fieldmanager is tested down to and works on PHP 5.3-5.6". I'm willing to break with that at some point, but only for a major release, and I don't want to hold this for "2.0".

```suggestion
	abstract public function render_content( $content );
```

php/class-fieldmanager-markdown.php
```diff
@@ -0,0 +1,31 @@
...
+	 */
+	public $field_class = 'markdown';
+
+	/**
+	 * Render content as html from parsed markdown.
+	 *
+	 * @param mixed $value Unused value.
+	 * @return string Rendered content.
+	 */
+	public function render_content( string $content ): string {
```
mboynes, line 28:
```suggestion
	public function render_content( $content ) {
```

php/class-fieldmanager-markup.php
```diff
@@ -0,0 +1,29 @@
...
+	 */
+	public $field_class = 'markup';
+
+	/**
+	 * Render content using `wp_kses_post()`.
+	 *
+	 * @param mixed $value Unused value.
+	 * @return string Rendered content.
+	 */
+	public function render_content( string $content = '' ): string {
```
mboynes, line 26:
```suggestion
	public function render_content( $content = '' ) {
```

php/class-fieldmanager-plaintext.php
```diff
@@ -0,0 +1,29 @@
...
+	 */
+	public $field_class = 'plaintext';
+
+	/**
+	 * Render content.
+	 *
+	 * @param mixed $value Unused value.
+	 * @return string Rendered content.
+	 */
+	public function render_content( string $content = '' ): string {
```
mboynes, line 26:
```suggestion
	public function render_content( $content = '' ) {
```

mboynes (commented):
A few more changes then it looks good to me to ship!

fieldmanager.php
jameswburke: Not a problem, updated.

php/class-fieldmanager-content.php
```diff
@@ -0,0 +1,51 @@
+<?php
+/**
+ * Abstract class file for Fieldmanager_Content
+ *
+ * @package Fieldmanager
+ */
+
+/**
+ * Static markdown field.
```
dlh01, line 9: Needs updating.

php/class-fieldmanager-content.php
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * Do not save this field. This class is purely informational.
+	 *
+	 * @var bool
+	 */
+	public $skip_save = true;
+
+	/**
+	 * Render content.
```
dlh01, line 35: This seems copied from the below method (or vice versa).

php/class-fieldmanager-markdown.php
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Class file for Fieldmanager_Markdown
+ *
+ * @package Fieldmanager
+ */
+
+use Fieldmanager\Libraries\Parsedown;
+
+/**
+ * Static markdown field.
```
dlh01, line 11:
```suggestion
 * Static Markdown field.
```

php/class-fieldmanager-markdown.php
```diff
@@ -0,0 +1,31 @@
...
+
+	/**
+	 * Override field_class.
+	 *
+	 * @var string
+	 */
+	public $field_class = 'markdown';
+
+	/**
+	 * Render content as html from parsed markdown.
```
dlh01, line 23:
```suggestion
	 * Render content as HTML from parsed Markdown.
```

php/class-fieldmanager-markdown.php
```diff
@@ -0,0 +1,31 @@
...
+	 *
+	 * @param mixed $value Unused value.
+	 * @return string Rendered content.
+	 */
+	public function render_content( $content ) {
+		return ( new Parsedown\Parsedown() )->text( $content );
```
dlh01, line 29: Is it necessary, or preferable, to create a `Parsedown` each time `render_content()` is called?

dlh01, line 29: Also, should the value go through `wp_kses_post()`?

dlh01 (approved):
This looks good to me, too. I've noted a few documentation issues to tidy up, along with a couple of other suggestions.

I'm still a little hesitant about some of the class naming, although I think we can still merge this and revisit before release. 

`Fieldmanager_Markup` is ambiguous to me; Matt gave the example of a child class that renders Textile, and Textile is also a markup language.

Also, leaving out the `_Content_` prefix from the child class names is potentially cutting us off from the obvious name of a future field that is designed to actually save Markdown, or whatever. The context classes are `Fieldmanager_Context_*`; the datasource classes are `Fieldmanager_Datasource_*`; maybe these should be `Fieldmanager_Content_*`.

php/class-fieldmanager-markdown.php
jameswburke: Good thinking, found a static `instance()` method and modified to use that.

jameswburke: I'm still a bit on the fence, but I think it's for the best.

php/class-fieldmanager-plaintext.php
```diff
@@ -0,0 +1,29 @@
...
+	public $field_class = 'plaintext';
+
+	/**
+	 * Render content.
+	 *
+	 * @param mixed $value Unused value.
+	 * @return string Rendered content.
+	 */
+	public function render_content( $content = '' ) {
+		return esc_html( $content );
```
dlh01, line 27: Late-breaking thought: Should we run this through `wpautop()` to ensure that it's not text sitting in a `<div>`?

jameswburke: I don't have strong feelings on this either way. Are you throwing the idea out there? Or do you think we should go ahead and do it?

dlh01: My only goal is to ensure the content is wrapped in something semantic. Beyond that, I'm not picky. Without checking out the branch at this moment, if the content is currently within a `div` and using `wpautop` will make it not be in a `div`, then I'd take it. Maybe wrapping the return value in `<p>` and `</p>` could also do it.

dlh01: I had a chance to check out the branch and see that the content currently displays directly within `div.fm-item.fm-plaintext`. It seems reasonable to me to assume that a paragraph will accurately describe the kind of content that would be included with the plaintext field (with complex content going into the Markdown or HTML fields). For that reason, running the output through `wpautop()` would be my recommendation.

jameswburke: I just pushed up a commit introducing `wpautop` to the plaintext field.

php/libraries/parsedown/composer.json
```diff
@@ -0,0 +1,33 @@
+{
```
mboynes, line 1: You can delete this file

php/libraries/parsedown/README.md
```diff
@@ -0,0 +1,103 @@
+<!-- ![Parsedown](https://i.imgur.com/yE8afYV.png) -->
```
mboynes, line 1: I'm inclined to say delete this file too, but I can be convinced it should stay if anyone feels strongly about it

PR #785, opened 2020-11-17
Fieldmanager Webpack setup for JS and CSS minification

stevenslack:
This PR should serve as a springboard to using Webpack as a build system. This PR has the following characteristics:
* Minification of JS files into the `build` directory
* Minification of the CSS files into the build directory
* Apply `autoprefixer` to CSS properties using PostCSS
* Address existing NPM audit vulnerabilities

Future efforts may want to include:
- style linting
- Javascript linting
- JS Unit tests with Jest
- SASS integration

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,23 @@
   },
   "license": "GPL-2.0",
   "devDependencies": {
-    "grunt": "^0.4.5",
+    "css-loader": "^5.0.1",
+    "css-minimizer-webpack-plugin": "^1.1.5",
+    "grunt": "^1.3.0",
     "grunt-contrib-connect": "~0.11.0",
-    "grunt-contrib-qunit": "~0.7.0",
-    "grunt-phpcs": "^0.4.0"
+    "grunt-contrib-qunit": "^4.0.0",
+    "grunt-phpcs": "^0.4.0",
+    "mini-css-extract-plugin": "^1.3.1",
+    "webpack-cli": "^4.2.0"
+  },
+  "scripts": {
+    "build": "webpack --mode=production"
+  },
+  "dependencies": {
+    "clean-webpack-plugin": "^3.0.0",
+    "postcss": "^8.1.7",
+    "postcss-loader": "^4.0.4",
+    "postcss-preset-env": "^6.7.0",
+    "webpack": "^5.4.0"
   }
 }
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,99 @@
+const path = require('path');
+const webpack = require('webpack');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
+const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+
+module.exports = (env, argv) => {
+  const { mode } = argv;
+
+  return {
+    devtool: mode === 'production'
+      ? 'source-map'
+      : 'cheap-module-eval-source-map',
+    entry: {
+      'fieldmanager': [
+        './js/fieldmanager.js',
+        './css/fieldmanager.css',
+        './css/fieldmanager-options.css',
+      ],
+      'fieldmanager-autocomplete': './js/fieldmanager-autocomplete.js',
+      'fieldmanager-colorpicker': './js/fieldmanager-colorpicker.js',
+      'fieldmanager-datepicker': './js/fieldmanager-datepicker.js',
+      'fieldmanager-draggablepost': [
+        './js/fieldmanager-draggablepost.js',
+        './css/fieldmanager-draggablepost.css',
+      ],
+      'fieldmanager-group-tabs': [
+        './js/fieldmanager-group-tabs.js',
+        './css/fieldmanager-group-tabs.css',
+      ],
+      'fieldmanager-media': './js/media/fieldmanager-media.js',
+      'fieldmanager-select': './js/fieldmanager-select.js',
+      'fieldmanager-quickedit': './js/fieldmanager-quickedit.js',
+      'fieldmanager-options': './css/fieldmanager-options.css',
+      'richtext': './js/richtext.js',
+    },
+    output: {
+      filename: mode === 'production'
+        ? 'js/[name].bundle.min.js'
+        : 'js/[name].js',
+      path: path.join(__dirname, 'build'),
+    },
+    module: {
+      rules: [
+        {
+          test: /.css$/,
+          use: [
+            MiniCssExtractPlugin.loader,
+            'css-loader',
+            {
+              loader: 'postcss-loader',
+              options: {
+                postcssOptions: {
+                  plugins: [
+                    [
+                      // postcss-preset-env includes autoprefixer
+                      'postcss-preset-env',
+                      {
+                        browsers: 'last 2 versions',
+                      },
+                    ],
+                  ],
+                },
+              },
+            },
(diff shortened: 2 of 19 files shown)


webpack.config.js
```diff
@@ -0,0 +1,99 @@
...
+            {
+              loader: 'postcss-loader',
+              options: {
+                postcssOptions: {
+                  plugins: [
+                    [
+                      // postcss-preset-env includes autoprefixer
+                      'postcss-preset-env',
+                      {
+                        browsers: 'last 2 versions',
```
stevenslack, line 59: This may need to be adjusted but for the most cases the -webkit and -moz prefixes are added here with autoprefixer

css/fieldmanager-richtext-content.css
```diff
@@ -1,144 +0,0 @@
...
-	height: 250px;
-}
-
-img.wp-oembed {
-	border: 1px dashed #888;
-	background: #f7f5f2 url("img/embedded.png") no-repeat scroll center center;
-	width: 300px;
-	height: 250px;
-}
-
```
montchr, line 144: Is this change intentional? I would expect these build workflow changes to leave existing styles in place (aside from the vendor prefix removals).

webpack.config.js
```diff
@@ -0,0 +1,99 @@
...
+      ],
+      'fieldmanager-group-tabs': [
+        './js/fieldmanager-group-tabs.js',
+        './css/fieldmanager-group-tabs.css',
+      ],
+      'fieldmanager-media': './js/media/fieldmanager-media.js',
+      'fieldmanager-select': './js/fieldmanager-select.js',
+      'fieldmanager-quickedit': './js/fieldmanager-quickedit.js',
+      'fieldmanager-options': './css/fieldmanager-options.css',
+      'richtext': './js/richtext.js',
```
montchr, line 35: Could these files be organized into module/entry-specific directories? Doing so might make the bundle structure more intuitive at a glance, as the file/directory structure could indicate which files are bundled together.

I know we may want to retain the `fieldmanager-*` naming convention for the built assets, but now that we're compiling them via Webpack, perhaps the `fieldmanager-` prefix could be dropped from the source files?

webpack.config.js
```diff
@@ -0,0 +1,99 @@
...
+        '...', // Load existing minimizers in order to minimize JS with terser-webpack-plugin.
+        new CssMinimizerPlugin(),
+      ],
+    },
+    resolve: {
+      extensions: ['.js', '.css'],
+    },
+    externals: {
+      // Enable require('jquery') where jquery is already a global
+      jquery: 'jQuery',
```
montchr, line 96: Could you explain this a little more? I'm not seeing `require('jquery')` in use anywhere. I may be misunderstanding its intention but wouldn't that result in jQuery being bundled in the built assets in addition to loading it as a dependency via the enqueue system?

webpack.config.js
montchr: Consider adding this to `package.json` via the `browserslist` field (or a `.browserslistrc` file) instead so other tools like Babel can make use of it too. 🙂 

https://github.com/csstools/postcss-preset-env#browsers
https://babeljs.io/docs/en/babel-preset-env

build/js/fieldmanager.bundle.min.js
```diff
@@ -0,0 +1,2 @@
...d","false"):t(this).attr("aria-expanded","true")})),t(".fm-collapsed > .fm-group:not(.fmjs-proto) > .fm-group-inner").hide(),m.init_display_if=function(){var i,e=t(this).data("display-src"),a=l(this),r=t(this).siblings(".fm-"+e+"-wrapper").find(".fm-element").not("div, .fm-autocomplete, .fm-checkbox-hidden");0!==r.length&&(i=r.is(":checkbox")?r.is(":checked")?r.val():r.siblings('input[type=hidden][name="'+r.attr("name")+'"]').val():r.is(":radio")?r.filter(":checked").length?r.filter(":checked").val():"":r.val().split(","),r.addClass("display-trigger"),o(a,i)||t(this).hide())},t(".fm-display-if").each(m.init_display_if),m.trigger_display_if=function(){var i,e=t(this),a=e.attr("name");i=e.is(":checkbox")?e.is(":checked")?e.val():e.siblings('input[type=hidden][name="'+a+'"]').val():e.is(":radio")?e.filter(":checked").val():e.val().split(","),t(this).closest(".fm-wrapper").siblings().each((function(){t(this).hasClass("fm-display-if")&&a&&null!=a.match(t(this).data("display-src"))&&(o(l(this),i)?t(this).show():t(this).hide(),t(this).trigger("fm_displayif_toggle"))}))},t(document).on("change",".display-trigger",m.trigger_display_if),s(),r(),t(document).on("fm_activate_tab",r)}))})()})();
```
montchr, line 1: It looks like the built file uses arrow functions, which aren't supported in IE. Should we bring in Babel as a dev dependency to ensure compatibility with supported browsers?

montchr (changes requested):
Awesome stuff! 😄 Just a few questions and suggestions.

css/fieldmanager-richtext-content.css
stevenslack: Yes, as far as I could tell this was a stale file and was not being used.

build/js/fieldmanager.bundle.min.js
stevenslack: Good catch!

Perhaps we should bring in Babel for this reason.

webpack.config.js
stevenslack: @montchr Yes, I considered this as well. The intention was to keep the changes minimal at best. But considering some of your findings here it may best serve to restructure for organization purposes. Additionally, if we were to Sassify the CSS the restructuring would benefit those efforts

package.json
```diff
@@ -8,9 +8,24 @@
...
+  "scripts": {
+    "build": "webpack --mode=production",
+    "grunt": "grunt"
+  },
+  "dependencies": {
+    "clean-webpack-plugin": "^3.0.0",
+    "postcss": "^8.1.7",
+    "postcss-loader": "^4.0.4",
+    "postcss-preset-env": "^6.7.0",
+    "webpack": "^5.4.0"
```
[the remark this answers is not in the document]
stevenslack: Yes, the build runs successfully on the current LTS `14.15.1`.

webpack.config.js
stevenslack: The `externals` method assumes that jQuery is a global in the consumer's environment, WP in this case. If this is not provided then the build will error as it will look to resolve jQuery in the project as a dependancy. 

> The externals configuration option provides a way of excluding dependencies from the output bundles. Instead, the created bundle relies on that dependency to be present in the consumer's (any end-user application) environment.

-- https://webpack.js.org/configuration/externals/

build/js/fieldmanager.bundle.min.js
stevenslack: Webpack 5 uses targets for determining how to compile the code. If there is a `browserslist` config in the package.json file it will use that. I tested and it removed the arrow functions.
https://webpack.js.org/configuration/target/#browserslist

stevenslack: https://webpack.js.org/blog/2020-10-10-webpack-5-release/#improved-target

srtfisher:
closing due to age. if we were revive this I would want to see the webpack configuration based on https://github.com/alleyinteractive/create-wordpress-plugin now

Issue #621, opened 2017-06-04
Fieldmanager_Datasource::get_ajax_action() should also account for classname


mboynes:
Two distinct extensions of `Fieldmanager_Datasource` with identical `$options` will generate the same ajax key in `get_ajax_action()`. `get_ajax_action()` should include `get_called_class()` in the concatenated string it uses for the ajax hash.

dlh01:
Done in #794.

PR #794, opened 2021-02-14
Adds the calling class to the string for the ajax hash to prevent collisions

mslinnea:
Addresses https://github.com/alleyinteractive/wordpress-fieldmanager/issues/621

### DIFF
--- a/php/datasource/class-fieldmanager-datasource-post.php
+++ b/php/datasource/class-fieldmanager-datasource-post.php
@@ -193,7 +193,8 @@ class Fieldmanager_Datasource_Post extends Fieldmanager_Datasource {
 		}
 		$unique_key  = json_encode( $this->query_args );
 		$unique_key .= (string) $this->query_callback;
-		return get_called_class() . crc32( $unique_key );
+		$unique_key .= get_called_class();
+		return 'fm_datasource_post_' . crc32( $unique_key );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


dlh01 (approved):
🌳 I wondered whether the post, term, and user datasources should get the same fix, since they override the default `get_ajax_action()`. But I see that each has a distinct leading string like `fm_datasource_post` and `fm_datasource_term`...except for the user datasource, which also uses `fm_datasource_post` (oops!). Can we switch that to `fm_datasource_user` while we're here? (Or drop the `_type_` prefix from each and use `get_called_class()` like in the parent class.)

mboynes:
> except for the user datasource, which also uses `fm_datasource_post`

😆

mslinnea:
@dlh01 I updated the other classes to all use `get_called_class()` since that will prevent this same issue if those classes were extended.

dlh01:
@mslinnea Unfortunately, it seems that using `get_called_class()` on its own will cause datasources to break when the class name is under a namespace: The action name in `$_POST` will get slashed and will no longer match the action name generated by Fieldmanager, like `\\My\\Datasource` instead of `\My\Datasource`.

I'm wondering whether the class name should be part of the hashed string instead, so maybe `fm_datasource_{$hash}`.

mslinnea:
@dlh01 Valid point, I've updated these so that the called class is part of the hash, and I also fixed the preface for the user datasource.

dlh01 (approved):
🌳 Now we're cookin' with `get_called_class()`!

Issue #789, opened 2021-01-18
Fieldmanager_Grid using out of date jQuery and not persisting


jjrdevspace01:
tried to use Fieldmanager_Grid for the first time, i've had no problems with the other fieldmanager classes
followed the demo here https://github.com/alleyinteractive/fieldmanager-demos/blob/master/lib/class-fm-demo-grid.php

            'table' => new Fieldmanager_Grid( array(
                'js_options' => array(
                    'startRows'    => 6,
                    'startCols'    => 5,
                    'colHeaders'   => true,
                    'rowHeaders'   => true,
                    'minSpareCols' => 1,
                    'minSpareRows' => 1,
                    'contextMenu'  => true,
                )
            )),

First problem was 'grid.js?ver=1.2.6:88 Uncaught TypeError: $(...).live is not a function

using WP5.6 - Managed to work around that installing the plugin 'enable-jquery-migrate-helper'

However although this allowed the grid to display correctly i was unable to use the control to persist data to the post_meta

dlh01:
It might be too much to ask to deliver a fix for this issue in the current 1.4.0 cycle, but let's see whether we can at least diagnose it and determine a plan.

dlh01:
Hey @jjrdevspace01,

Thanks for filing this report. I'm sorry to say that we're not going to be able to provide a fix for this issue in the core Fieldmanager repo. We're instead opting to deprecate the current grid field — see #795.

That said, if a grid-like field is important for your projects, we wholeheartedly encourage you to spin off a custom extension to Fieldmanager in your own repo. Fieldmanager has a long history of being the base library for custom field classes that serve specific needs — for example, https://github.com/dlh01/fieldmanager-beta-customize or https://github.com/alleyinteractive/fm-zones. 

If you end up going this route, be sure to let us know where to find your work!

jjrdevspace01:
thanks @dlh01 - it was my first time trying the grid control so not much of an issue. I had a requirement to display a bunch of small table elements on a page and it looked like a good fit. i used a combination of fieldmanager groups and text based fields instead which is my usual approach to this. it works but is a little cumbersome editing and developing. It does feel like a grid type solution was a great idea in theory. If it was a bigger requirement something like tablepress and its shortcodes in richtext would work but for smaller needs a simple grid was a good idea even if the execution needed some refinement so is a shame it was dropped from the core product.

dlh01:
[...]

PR #795, opened 2021-02-16
Deprecate `Fieldmanager_Grid`

dlh01:
Grid has received little attention since its introduction and today appears to be nonfunctional -- see #789. Additionally, Handsontable, on which it is based, is now a commercial product. We can no longer recommend it and so now move to deprecate it with intent to remove it from the repository.


kjbenk (approved):
⛳

PR #796, opened 2021-02-19
DateField/AutoComplete Gutenberg Patch

### LINKED ISSUE #793: Fieldmanager_Datepicker not working in WordPress 5.6 on Gutenberg Post
The Fieldmanager_Datepicker JS isn't working in WordPress 5.6 on the Gutenberg edit post screen, but it does work fine in WordPress 5.5.2.

It also works fine in WordPress 5.6 for the classic editor.

Code to reproduce the issue:

```
add_action( 'fm_post', function () {
    $fm_date = new Fieldmanager_Datepicker( 'Date' );
    $fm_date->add_meta_box( 'Date', 'post' );
} );
```

jomurgel:
Closes #793.

- Allows Autocomplete and Date fields in Gutenberg to load jQuery scripts on input focus. In the Gutenberg context fields are not enabled as the jQuery `$( 'FIELD_NAME:visible' )` returns undefined, unless re-initialized as needed which prevents the need to auto-enabled ALL fields on the page (many of which are hidden by default using the `display_if` option).

### DIFF
--- a/js/fieldmanager-datepicker.js
+++ b/js/fieldmanager-datepicker.js
@@ -1,10 +1,12 @@
-( function( $ ) {
+(function( $ ) {
 	fm.datepicker = {
 		add_datepicker: function() {
 			$( '.fm-datepicker-popup:visible' ).each( function() {
-				if ( !$( this ).hasClass( 'fm-has-date-picker' ) ) {
-					var opts = $( this ).data( 'datepicker-opts' );
-					$( this ).datepicker( opts ).addClass( 'fm-has-date-picker' );
+				var $el = $( this );
+
+				if ( ! $el.hasClass( 'fm-has-date-picker' ) ) {
+					var opts = $el.data( 'datepicker-opts' );
+					$el.datepicker( opts ).addClass( 'fm-has-date-picker' );
 				}
 			} );
 		}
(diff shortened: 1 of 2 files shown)


isGabe (approved):
🐴

js/fieldmanager-datepicker.js
```diff
@@ -13,4 +13,8 @@
 	fmLoadModule( fm.datepicker.add_datepicker );
 
 	$( document ).on( 'fm_collapsible_toggle fm_added_element fm_displayif_toggle fm_activate_tab', fm.datepicker.add_datepicker );
+	$( document ).on( 'focus',
+			'input[class*="fm-datepicker-group"]:not(.fm-has-date-picker)',
+			fm.autocomplete.enable_autocomplete
+		);
```
dlh01, line 19: - The indentation of the last three lines seems inconsistent.
- `add_datepicker`, not `enable_autocomplete`?

dlh01:
@jomurgel Thanks! 

I found that a couple more adjustments were necessary: targeting the class used directly on the `input` element rather than a potential group class, and using `datepicker.add_datepicker` rather than `autocomplete.add_datepicker`. I've made these changes in https://github.com/alleyinteractive/wordpress-fieldmanager/pull/796/commits/bddfdf08eb8ad46ed3cbd8b06564c7a4ba9ee1cc.

For both the datepicker and autocomplete updates, I also removed the initialization bound to various Fieldmanager events. I don't think those will be necessary now that the focus event triggers the same behavior.

Would you mind testing the updates?

mslinnea (approved):
🐴  Tested this on 2 different sites. It resolves issues with date picker and autocomplete in WP 5.6

PR #801, opened 2021-03-18
Pass previous value when updating post meta in Storable context

### LINKED ISSUE #800: class-fieldmanager-context-storable.php doesn't pass previous value on update.
Ran into a bug today where previous value is never set when updating post meta.

Tracked the culprit down to this line:

https://github.com/alleyinteractive/wordpress-fieldmanager/blob/master/php/context/class-fieldmanager-context-storable.php#L90

Fix appears to be passing `$current` to `update_data`

PR incoming

anubisthejackle:
Closes #800 

Passes `$current` to `update_data` so `$prev_value` is available down the line.

### DIFF
--- a/php/context/class-fieldmanager-context-storable.php
+++ b/php/context/class-fieldmanager-context-storable.php
@@ -87,7 +87,7 @@ abstract class Fieldmanager_Context_Storable extends Fieldmanager_Context {
 		$data    = $this->prepare_data( $current, $data, $field );
 		if ( ! $field->skip_save ) {
 			if ( $field->serialize_data ) {
-				$this->update_data( $this->fm->data_id, $field->get_element_key(), $data );
+				$this->update_data( $this->fm->data_id, $field->get_element_key(), $data, $current );
 			} else {
 				$this->delete_data( $this->fm->data_id, $field->get_element_key() );
 				foreach ( $data as $value ) {


dlh01:
@anubisthejackle I'm not sure that this change is correct. Somewhat confusingly, `$prev_value` is not "the previous value for the field" but "update the data only if the previous value was this," following the metadata API:

https://github.com/alleyinteractive/wordpress-fieldmanager/blob/1994637fbe962b94c943d0dab591ab75609f605e/php/context/class-fieldmanager-context-storable.php#L222-L224

https://github.com/WordPress/wordpress-develop/blob/4d66791327ce485dc31c62692690e068eb150af8/src/wp-includes/meta.php#L151-L153

What do you think?

anubisthejackle:
That's true. The core problem is that $prev_value isn't available in update_post_metadata when it's expected. Maybe the real fix would be providing a flag somewhere to tell Fieldmanager to pass that along, or to not pass that along as it does currently.

dlh01:
> The core problem is that $prev_value isn't available in update_post_metadata when it's expected. 

I guess I'm stuck on when it would be expected, though?

> Maybe the real fix would be providing a flag somewhere to tell Fieldmanager to pass that along, or to not pass that along as it does currently.

Yeah, that might be a useful new feature.

PR #802, opened 2021-03-20
Allow end user to define a label for the empty option in a select

anubisthejackle:
Exactly as it says. This PR adds a parameter "empty_string" that allows changing of the displayed name for empty options. The value of this option remains empty.

### DIFF
--- a/php/class-fieldmanager-select.php
+++ b/php/class-fieldmanager-select.php
@@ -37,6 +37,13 @@ class Fieldmanager_Select extends Fieldmanager_Options {
 	public $first_empty = false;
 
 	/**
+	 * String to use for first empty element.
+	 *
+	 * @var string
+	 */
+	public $empty_string = '&nbsp;';
+
+	/**
 	 * Tell FM to save multiple values.
 	 *
 	 * @var bool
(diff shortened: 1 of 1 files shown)


renatonascalves (approved):
🗡️

dlh01:
This seems useful to me. A unit test for this behavior would be helpful alongside the other `first_empty`-related tests: https://github.com/alleyinteractive/wordpress-fieldmanager/blob/1994637fbe962b94c943d0dab591ab75609f605e/tests/php/test-fieldmanager-select-field.php. I would also offer that it might be beneficial for the property to be `[something]_label`, like `first_empty_label`, for greater consistency with other field properties like `add_more_label` and `button_label`.

anubisthejackle:
@dlh01 You're right. I've updated the label name and added a test for this.

php/class-fieldmanager-select.php
```diff
@@ -120,7 +127,7 @@ public function form_element( $value = array() ) {
 
 		$opts = '';
 		if ( $this->is_repeatable() || $this->first_empty ) {
-			$opts .= '<option value="">&nbsp;</option>';
+			$opts .= '<option value="">' . $this->first_empty_label . '</option>';
```
mboynes, line 130: this should be escaped:
```suggestion
			$opts .= '<option value="">' . esc_html( $this->first_empty_label ) . '</option>';
```

PR #686, opened 2018-01-12
Bugfix for saving empty multi selects or radio fields.

discoinfiltrator:
A few fixes here:

- $data should be an array, as that's how it's used in save_walk_children
- save_walk_children only has two function parameters, so removing $this->fm->data_id from the call
- Since we're relying on the POST array, we need to check our fields for multi-selects and radio buttons (which when empty are not passed in the POST array)

Relevant GitHub issue is:

https://github.com/alleyinteractive/wordpress-fieldmanager/issues/685

PR #811, opened 2021-11-07
Handle options fields in groups missing from `$_POST`

### LINKED ISSUE #685: Cannot save empty multi-select when serialize_data is false
I am using the multiselect (w/ chosen) to implement taxonomy term selection for posts:

```
$tax = new Fieldmanager_Select(
    [
        'label' => 'My Taxonomy',
        'type_ahead' => true,
        'multiple' => true,
        'datasource' => new Fieldmanager_Datasource_Term(
            [
                'taxonomy' => ['my_taxonomy_name'],
                'only_save_to_taxonomy' => true,
                'taxonomy_hierarchical' => true,
            ]
        )
    ]
);
$fm = new Fieldmanager_Group([
    'name' => 'my_group_name',
    'add_to_prefix' => false,
    'serialize_data' => false,
    'description_after_element' => false,
    'children' => [
        'my_taxonomy_name' => $tax,
    ],
]);
$fm->add_meta_box('Taxonomies (required)', ['post']);
```

I was running into a bug:

1. Add a taxonomy to a post, save.
2. Edit post, remove taxonomy, save.
3. Notice that post is saved, but taxonomy was not removed.

After digging through the code, the cause was that when serialize_data is true, we execute this code in the Fieldmanager_Context_Storable, save:

```
$data = isset( $_POST[ $this->fm->name ] ) ? $_POST[ $this->fm->name ] : '';
```

This is an issue for multi-selects and radio buttons.  Multiple select and radio fields with no values chosen are left out of the post request altogether. I've come up with a solution and will submit a PR.

### LINKED ISSUE #764: Can't uncheck all checkboxes if serialize_data set to false
In a `Fieldmanager_Checkboxes` input, if `serialize_data` is set to `false`, you cannot uncheck all the checkboxes and update a previously saved value.

If you have some options checked, save it, then uncheck them all, it doesn't register that you want to uncheck them all, and stays at the last setting.

Here's a code example:

```php
add_action( 'fm_post_post', function() {
	( new Fieldmanager_Group( [
		'name'           => 'test',
		'serialize_data' => false,
		'children'       => [
			'checkboxes' => new Fieldmanager_Checkboxes(
				'Checkboxes',
				[ 'options' => [ 'One', 'Two', 'Three' ] ]
			),
		]
	] ) )->add_meta_box( 'Testing Serialize Data with Checkboxes', 'post' );
} );
```

dlh01:
Refines #686, the original Pull Request addressing the original report in #685, and expands it to address the same issue with checkboxes as reported in #764.

The differences between this and #686:

- Apply the fix only when `false === $serialize_data`, since both of the reports specify that the bug applies to those fields. Also, otherwise, we could inadvertently save a value for `$data` that only contained a few of the fields in a  group.

- Check whether the field exists in `$data` with `array_key_exists()` rather than `isset()`. The idea is that this expands backwards compatibility by contracting the number of cases when the new fallback value will be added to `$data`.

- Define `$data = array()` only when the field is a group, again to expand backwards compatibility by contracting the number of cases when `$data` defaults to something new. Arguably, repeatable single fields could also default to `array()`, but they don't currently, and it doesn't seem necessary to change that to fix this issue.

- Don't provide a fallback for `Fieldmanager_Radios`. First, I would think the fallback should be scalar, not an array, since radios don't save multiple values. However, whereas checkboxes or multi-selects missing from `$_POST` could be interpreted as an empty set, a radio input is missing from `$_POST` means that not one of the options was selected — wouldn't it be incorrect to supply a default? `''` might well have been one of the available values that wasn't selected by the user. And since Fieldmanager doesn't include a way to un-select a radio option, it seems like the bug in #686 would apply only before the user makes a selection and that that period of time should be respected. 

- Don't extend the new behavior to subclasses, which could already supply their own workarounds or alternate behavior.

Closes #685.
Closes #764.

### DIFF
--- a/php/context/class-fieldmanager-context-storable.php
+++ b/php/context/class-fieldmanager-context-storable.php
@@ -49,13 +49,45 @@ abstract class Fieldmanager_Context_Storable extends Fieldmanager_Context {
 
 		$this->fm->current_context = $this;
 		if ( $this->fm->serialize_data ) {
-			$this->save_field( $this->fm, $data, $this->fm->data_id );
+			$this->save_field( $this->fm, $data );
 		} else {
 			if ( null === $data ) {
-				$data = isset( $_POST[ $this->fm->name ] ) ? $_POST[ $this->fm->name ] : ''; // WPCS: input var okay. CSRF okay. Sanitization okay.
+				if ( isset( $_POST[ $this->fm->name ] ) ) { // WPCS: input var okay. CSRF okay. Sanitization okay.
+					$data = $_POST[ $this->fm->name ]; // WPCS: input var okay. CSRF okay. Sanitization okay.
+				} elseif ( $this->fm->is_group() ) {
+					$data = array();
+				} else {
+					$data = '';
+				}
+
+				if ( false === $this->fm->serialize_data && $this->fm->is_group() ) {
+					/*
+					 * Hardcoded exceptions for scenarios involving bundled
+					 * fields that are omitted from $_POST when empty.
+					 *
+					 * @see https://github.com/alleyinteractive/wordpress-fieldmanager/issues/685
+					 * @see https://github.com/alleyinteractive/wordpress-fieldmanager/issues/764
+					 */
+					foreach ( $this->fm->children as $child ) {
+						if ( array_key_exists( $child->name, $data ) ) {
+							continue;
+						}
+
+						$type = get_class( $child );
+
+						if (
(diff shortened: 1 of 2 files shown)


mogmarsh (approved):
👍  looks good. happy to have my bug resolved, although the need is long gone

mslinnea (approved):
Looks good and solid reasoning 👍 🐴

discoinfiltrator:
Looks great 👍

PR #815, opened 2021-11-29
Fix FM module loader in Gutenberg context

kjbenk:
# Problem
Most FM initializes happens when the DOM is ready and only if that element is considered visible via the jQuery `.is(':visible')` function. In a recent Gutenberg update, metabox contents are not rendered until after the DOM is ready. (related to - https://github.com/alleyinteractive/wordpress-fieldmanager/issues/793#issuecomment-778538322)

This can lead to issues with initialization since some metabox elements will not be rendered properly on DOM ready and therefore will return `false` for `.is(':visible')`.

Potentially closes - https://github.com/alleyinteractive/wordpress-fieldmanager/pull/812

# Solution
When loading FM in a Gutenberg context, hook into the `areMetaBoxesInitialized` (wait a short period of time) and perform all FM initializations. This ensures that all metabox elements are rendered and in the DOM before any `.is(':visible')` checks are executed.

### DIFF
--- a/fieldmanager.php
+++ b/fieldmanager.php
@@ -18,7 +18,7 @@ Author URI: https://www.alley.co/
 /**
  * Current version of Fieldmanager.
  */
-define( 'FM_VERSION', '1.3.0' );
+define( 'FM_VERSION', '1.3.1' );
 
 /**
  * Filesystem path to Fieldmanager.
--- a/js/fieldmanager-loader.js
+++ b/js/fieldmanager-loader.js
@@ -4,22 +4,44 @@
  * Ensures that metaboxes have loaded before initializing functionality.
  * @param {function} callback - The callback function to execute when the DOM is ready.
  */
-function fmLoadModule( callback ) {
-	if ( 'object' === typeof wp && 'function' === typeof wp.domReady ) {
-		wp.domReady( callback );
-	} else if ( jQuery ) {
-		jQuery( document ).ready( callback );
-	} else {
-		// Shim wp.domReady.
-		if (
-			document.readyState === 'complete' || // DOMContentLoaded + Images/Styles/etc loaded, so we call directly.
-			document.readyState === 'interactive' // DOMContentLoaded fires at this point, so we call directly.
-		) {
-			callback();
-			return;
-		}
+ function fmLoadModule( callback ) {
+	// For the Gutenberg editor only, load FM module after metabox initialization.
+	if (wp.data) {
+		var metaboxesInitializedUnsubscribe = wp.data.subscribe(() => {
+			if (wp.data.select( 'core/edit-post' ).areMetaBoxesInitialized()) {
+				/**
+				 * `areMetaBoxesInitialized` is called immediately before the
+				 * `MetaBoxesArea` component is rendered, which is where the metabox
+				 * HTML is moved from the hidden div and into the main form element.
+				 *
+				 * This means we need to delay the callback a bit to allow this
+				 * component to render such that the jQuery `:visible` selector
+				 * can work as expected.
+				 */
+				setTimeout(() => {
+					callback();
+				}, 100);
 
-		// DOMContentLoaded has not fired yet, delay callback until then.
-		document.addEventListener( 'DOMContentLoaded', callback );
+				metaboxesInitializedUnsubscribe();
+			}
+		});
+	} else {
+		if ( 'object' === typeof wp && 'function' === typeof wp.domReady ) {
+			wp.domReady( callback );
+		} else if ( jQuery ) {
+			jQuery( document ).ready( callback );
+		} else {
+			// Shim wp.domReady.
+			if (
+				document.readyState === 'complete' || // DOMContentLoaded + Images/Styles/etc loaded, so we call directly.
+				document.readyState === 'interactive' // DOMContentLoaded fires at this point, so we call directly.
+			) {
+				callback();
+				return;
+			}
+
+			// DOMContentLoaded has not fired yet, delay callback until then.
+			document.addEventListener( 'DOMContentLoaded', callback );
+		}
 	}
 }


dlh01:
@kjbenk So that folks can test more easily, could you post (or link to) example fields that are subject to the bug, and what the expected behavior is before and after the patch is applied?

kjbenk:
@dlh01 Thanks for the input :) I added some steps to reproduce the issue and the expected behavior.

js/fieldmanager-loader.js
```diff
@@ -4,22 +4,55 @@
...
+					 * This means we need to delay the callback a bit to allow this
+					 * component to render such that the jQuery `:visible` selector
+					 * can work as expected.
+					 *
+					 * @link https://github.com/WordPress/gutenberg/blob/019d0a1b1883a5c3e5c9cdecc60bd5e546b60a1b/packages/edit-post/src/components/meta-boxes/index.js#L38-L45
+					 * @link https://github.com/WordPress/gutenberg/blob/d39949a3b9dc8e12d5f5d33b9091f14b93b37c8a/packages/edit-post/src/components/meta-boxes/meta-boxes-area/index.js#L34-L36
+					 */
+					setTimeout(() => {
+						callback();
+					}, 100);
```
mboynes, line 31: There must be a way to do this without the `setTimeout` call, let's keep at it until we identify what that is. `setTimeout` is too fragile for this.

mboynes (changes requested):

js/fieldmanager-loader.js
kjbenk: Totally agree :) I figured starting here is better than nothing for the time being.

fieldmanager.php
```diff
@@ -18,7 +18,7 @@
 /**
  * Current version of Fieldmanager.
  */
-define( 'FM_VERSION', '1.3.0' );
+define( 'FM_VERSION', '1.3.1' );
```
dlh01, line 21: The `main` branch has had several changes merged since the release of `1.3.0`, so I don't think it'd be accurate to label it `1.3.1` after this is merged. I would suggest leaving this line at `1.3.0` for the purpose of this particular PR. 

If you think the fix should also be backported to create a new 1.3.1 release, perhaps we can do that by creating a `1.3` branch based on the `1.3.0` tag (https://github.com/alleyinteractive/wordpress-fieldmanager/releases/tag/1.3.0) and merging this change into that branch in a separate PR, also bumping the version.

js/fieldmanager-loader.js
kjbenk: I just did some quick testing and something like this works. The only thing I do not like about this approach is that we are not checking any state within the subscribe function, but instead checking a DOM element. Ideally, the code we run within `wp.data.subscribe` would validate against the store, however, this might be a good option for now.

```js
	const wrappedCallback = () => {
		if (document.querySelector('.block-editor-page')) {
			const metaboxesInitializedUnsubscribe = wp.data.subscribe(() => {
				// Metaboxes are rendered and in the DOM.
				if ( document.querySelector('.edit-post-meta-boxes-area__container') ) {
					callback();

					// Clean up.
					metaboxesInitializedUnsubscribe();
				}
			});
		} else{
			callback();
		}
	};
```

fieldmanager.php
kjbenk: Good point :) Removed the version update

js/fieldmanager-loader.js
joemcgill: I think subscribing to the `areMetaBoxesInitialized` state is the right approach, but rather than using a `setTimeout`, I think we could wrap the callback in a polling function that checks for the existence of the initialized markup before firing our callback. The polling function would need to have an upper limit of the amount of times it would run, but should work great.

joemcgill:
I took a stab at updating @kjbenk's previous approach with a polling function in 08926e6. I think this a good enhancement on the original `setTimeout` approach that is a bit less fragile but works similarly in that it sets a polling function once the wp.data store has an updated state that the meta boxes have been initialized, but waits to make sure the markup is in place before firing the callback.

js/fieldmanager-loader.js
```diff
@@ -4,22 +4,71 @@
...
-function fmLoadModule( callback ) {
+ function fmLoadModule( callback ) {
+	/**
+	 * Wraps the provided callback to add check for the Gutenberg editor. If this
+	 * is called within the Gutenberg editor, then the callback function will trigger
+	 * after the metaboxes are initialized.
+	 */
+	const wrappedCallback = () => {
+		if (document.querySelector('.block-editor-page')) {
+			let pollAttempts = 0;
```
kevinfodness, line 15: Rather than setting this to `0` here and hardcoding the number of attempts it can't be greater than below (`5`) why not set this to `5` (and maybe rename it to something like `pollAttemptsRemaining`) and decrement it instead? Then the check could just be whether there are any poll attempts remaining. Currently, the threshold for polling attempts is pretty far down into the logic.

kevinfodness (approved):
Left a suggestion, but I like this a lot. 🍣

js/fieldmanager-loader.js
joemcgill: Good suggestion. Updated to decrement the counter in 1f4697f.

js/fieldmanager-loader.js
```diff
@@ -5,21 +5,70 @@
...
+						if ( document.querySelector('.edit-post-meta-boxes-area__container') ) {
+							callback();
+							callbackExecuted = true;
+						}
+
+						// Make sure we clear the callback interval.
+						if (callbackExecuted || ! pollAttemptsRemaining ) {
+							clearInterval(pollForMetaBoxes);
+						}
+					}, 100);
```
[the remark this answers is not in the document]
mboynes: yes, it looks like [subscribing a listener returns a function to unsubscribe it](https://github.com/WordPress/gutenberg/blob/9faa63ec01c7db4e4d57823af2db5c54bb6d6660/packages/data/src/utils/emitter.js#L23), as with redux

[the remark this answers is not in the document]
kjbenk: FWIW @joemcgill here is how I implemented the unsubscribe logic - https://github.com/alleyinteractive/wordpress-fieldmanager/pull/815/commits/c25c3cacea224dddc1ed3f00b572144191232a21#diff-92affb4a9da5be89b7b8cf47a3822c1eba297ee97c9f7487feaea5f214726c45R33

joemcgill:
@mboynes If we want to be in explicit control of when the callback gets fired, I think we'll need some way of handling the timing within our listener. That said, you're correct in the pragmatic sense in that...

> Further, is an interval even necessary? Once subscribed, won't this repeatedly get executed during ~any activity in the editor? And to that point, should this get properly unsubscribed once successfully executed?

We can probably just rely on state changes here instead of an interval and properly unsubscribe the listener after the callback is called. I've updated to that approach in 6f00450 and refined in 621cfa3. This is much simpler and works well.

mboynes (approved):

PR #816, opened 2021-12-06
Backport FM module loader fix to the 1.3 branch

joemcgill:
This backports the FM loader fix to the 1.3 branch.

Props @kjbenk.
See #815.


joemcgill:
This needs to be replaced with the updated approach after #815 lands.

PR #820, opened 2021-12-14
Loader: Fix loader callback timing

joemcgill:
This is a followup to 754f101e to make sure that we wait for the `.metabox-location-normal` form to be moved into `.edit-post-meta-boxes-area__container` in the block editor before running our loading callbacks. Otherwise, there is the possibility that the callbacks will still fire too early, before the FM forms are ready.


kjbenk (approved):
Nice ⛳

kjbenk (approved):

PR #821, opened 2022-01-06
Remove Grunt

dlh01:
The installed version of Grunt is very old, and it's unlikely that resurrected JavaScript unit tests will use Grunt to run.


stevenslack (approved):
This is great! Thanks @dlh01

Issue #825, opened 2022-02-09
Composer/Installers requirement out of date


crazyjaco:
I am hitting an issue pulling in fieldmanager via composer because of an incompatibility with the composer/installers requirement.

The fm repo has a semver of ~1.0.
The latest version of composer/installers is 2.0.1 (which we have in our composer file)
https://github.com/composer/installers/releases

Is it possible to raise the requirement to ^2.0 or *?
I realize * is not performant but I also don't want to negatively impact folks that are still on ^1.0 like our repo was until a week ago.

crazyjaco:
This was the error I am getting when trying to update the package.
`    - alleyinteractive/wordpress-fieldmanager 1.0.0 requires composer/installers ~1.0 -> found composer/installers[v1.0.0, ..., 1.x-dev] but it conflicts with your root composer.json require (2.0.1).`

crazyjaco:
composer/installers appears to be compatible with Composer 2.0 starting at version 1.9. If we changed ~1.0 to ^1.0, I think that would suffice. Will update the PR

theMikeD:
I am running into this now too. Evidently this has not yet been merged @dlh01? Or am I doing something wrong?

```
composer require alleyinteractive/wordpress-fieldmanager:^1.5
./composer.json has been updated
Running composer update alleyinteractive/wordpress-fieldmanager
Loading composer repositories with package information                                                                                                                                                                                              Updating dependencies                                 
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - composer/installers v1.0.22 requires composer-plugin-api 1.0.0 -> found composer-plugin-api[2.6.0] but it does not match the constraint.
    - composer/installers[v1.0.23, ..., v1.8.0] require composer-plugin-api ^1.0 -> found composer-plugin-api[2.6.0] but it does not match the constraint.
    - Root composer.json requires alleyinteractive/wordpress-fieldmanager ^1.5 -> satisfiable by alleyinteractive/wordpress-fieldmanager[1.5.0].
    - alleyinteractive/wordpress-fieldmanager 1.5.0 requires composer/installers ~1.0 -> satisfiable by composer/installers[v1.0.0, ..., 1.x-dev].
    - You can only install one version of a package, so only one of these can be installed: composer/installers[v1.0.0, ..., 1.x-dev, v2.0.0-alpha1, ..., 2.x-dev].
[...]

PR #826, opened 2022-02-09
Updating semver for composer/installers.

crazyjaco:
This is for issue #825.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     }
   ],
   "require": {
-    "composer/installers": "^2.0"
+    "composer/installers": "^1.0"
   },
   "require-dev": {
     "alleyinteractive/alley-coding-standards": "^0.3.0",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     }
   ],
   "require": {
-    "composer/installers": "^1.0"
+    "composer/installers": ">=1.0"
   },
   "require-dev": {
     "alleyinteractive/alley-coding-standards": "^0.3.0",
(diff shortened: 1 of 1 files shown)


dlh01:
Thanks for the PR, @crazyjaco!

Is there any reason to _not_ bump `composer/installers` to `^2.0`, as you first proposed in the issue?

crazyjaco:
The original constraint was ~1.0, which means 1.0, 1.0.1, 1.0.4 would be acceptable, but 1.2 or 1.4 would not.
Going to ^2.0 means that 2.0.2 and 2.5 would both be acceptable.

My concern going to 2.0 would force users to update other dependencies as well, but I guess thats not really this package's problem.

Honestly, it feels like the best solution is to remove the constraint entirely.
Thats what we did for most of our internal plugins.
The plugin/package works fine with most versions, if not all, so putting a constraint on it makes it difficult to maintain when a lot of plugins/packages are in play for a project.

crazyjaco:
It looks like ">=1.0" would also work. I didn't realize that was an option until I just looked again.

I use this tool to figure out semver impact
https://semver.npmjs.com/

dlh01:
Hey @crazyjaco — as you can see in the automated test results, the `>=` constraint fails Composer's `--strict` validation requirements. We might have to go with something like `^1.0 || ^2.0`.

crazyjaco:
Thanks! Updated with that suggestion. I didn't realize you could list a constraint like that. Neat!

dlh01 (approved):

Issue #829, opened 2022-03-01
fm_renumber performance


mboynes:
`fm_renumber()` in `fieldmanager.js` performs a lot of DOM traversal, and it would appear that much of it is duplicative in deeply nested groups and should be unnecessary.

This function should be refactored to, ideally, only visit each field once. Second best option is that each field is visited only the number of times as it has incrementable keys (as denoted by `data-fm-array-position` in `.fm-wrapper` nodes). As it presently stands, each field is visited an exponential number of times based on its depth and the ancestral group structure.

Also see #827, #828

PR #830, opened 2022-03-02
Limit fm_renumber traversal to repeatables

mboynes:
This is a performance improvement to fm_renumber. It works by limiting the `fm-wrapper` elements which get `data-fm-array-position` attributes to only those with limits != 1. In JS, only`.fm-element[data-fm-array-position]` elements are traversed. Using a sample group with deep nesting, DOM traversal dropped from thousands of elements to dozens and performance improved about as one would expect.

To be clear, this isn't an end-all-be-all *_fix_*. DOM traversal in this function is still a problem, and it's still very possible to generate a nesting of groups and fields that blocks the event loop for multiple seconds on sorting, but this should solve 99.9% of real world use cases. Further refinement should be weighed by its opportunity costs, since that time could be better spent on a deeper overhaul using React or VueJS.

Refs #829﻿

### DIFF
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -15,8 +15,7 @@ var init_sortable_container = function( el ) {
 				$( document ).trigger( 'fm_sortable_drag', el );
 			},
 			stop: function( e, ui ) {
-				var $parent = ui.item.parents( '.fm-wrapper' ).first();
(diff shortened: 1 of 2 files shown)


mogmarsh (approved):
👍

PR #838, opened 2022-06-14
adds action

mogmarsh:
This action allows other plugins such as https://github.com/alleyinteractive/fm-gutenberg to perform an action when a FieldManager context is constructed.


manewc (approved):
🎬

Issue #839, opened 2022-08-24
PHP 8.0: Autocomplete field with Extra Element causes Array to String Conversion Warning


mslinnea:
Fieldmanager Autocomplete will attempt to output an array as a default value when it should be a string.  This happens with a datasource like Fieldmanager_Datasource_Term when no results are returned.  The default is an empty array.  This can be reproduced with code that specifies one extra element and no default value.  For example, I've placed the following code in mu-plugins:

```
add_action( 'plugins_loaded', function () {
	$fm_sections = new Fieldmanager_Autocomplete(
		[
			'attributes'                => [
				'style' => 'width: 95%',
			],
			'exact_match'               => true,
			'remove_default_meta_boxes' => true,
			'name'                      => 'test_autocomplete',
			'limit'                     => 1,
			'extra_elements'            => 1,
			'add_more_label'            => 'Add Tag',
			'datasource'                => new Fieldmanager_Datasource_Term(
				[
					'taxonomy'              => 'post_tag',
					'append_taxonomy'       => false,
					'only_save_to_taxonomy' => true,
				]
			),
		]
	);
	$fm_sections->add_meta_box( 'Add Tag', 'post' );
} );
```

This triggers the following error:

<img width="1510" alt="Screen Shot 2022-08-23 at 5 21 35 PM" src="https://user-images.githubusercontent.com/7308162/186288251-e31b4501-9e3e-4893-8fca-6d5137ed3265.png">

PR #840, opened 2022-08-24
PHP 8.0: Autocomplete field with Extra Element causes Array to String Conversion Warning

mslinnea:
This resolves https://github.com/alleyinteractive/wordpress-fieldmanager/issues/839

I am adding a test along with this patch. Without this patch, the test will fail, as pictured:

Screen Shot 2022-08-23 at 5 28 43 PM


mogmarsh (approved):
👍

PR #854, opened 2023-02-15
Replacing set-output with GITHUB_OUTPUT, moving to shared PHPCS configuration

srtfisher:
- Replaces set-output with GITHUB_OUTPUT to fix the deprecation: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- Move the phpcs action to the shared action for the organization.


mogmarsh (approved):
👍

Issue #863, opened 2024-01-30
PHP Deprecated:  htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated


renatonascalves:
### Description

Currently, if one does not set a `default_value` field, it defaults to `null`, and a site returns this following PHP deprecation error.

```

[27-Jan-2024 18:07:59 UTC] PHP Deprecated:  htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/site/wp-includes/formatting.php on line 4715
```

The issue happens here: https://github.com/alleyinteractive/wordpress-fieldmanager/blob/608cbc18077e21e22533d91db709b6566cbc7903/php/class-fieldmanager-textarea.php#L50

Which calls https://github.com/WordPress/wordpress-develop/blob/94e6fe9f8ad05571d810c4a7fc2afdeeea1c0fee/src/wp-includes/formatting.php#L4732

### Use Case

A site with fields that extend `Fieldmanager_Field`:

```

add_action( 'fm_post_post', function() {
	$fm_title = new Fieldmanager_Textarea(
		[
			'name' => 'newsletter_excerpt',
			'description' => __( 'Excerpt to be included in json feed for use in newsletters', 'example-site' ),
		]
	);
	$fm_title->add_meta_box( __( 'Newsletter Excerpt', 'example-site' ), $post_types, 'normal', 'high' );
} );
```

### Acceptance Criteria

- Modify the logic that prints the field in WP admin so that if a default value is not set, and it returns `null`, then `esc_textarea` does not receive a value of `null` (either use a ternary to print an empty string, or pass an empty string to `esc_textarea` - do not modify the default value generally, since that could have other effects elsewhere that we want to avoid).

renatonascalves:
Maybe we just need to default to an empty string in the `Fieldmanager_Textarea` class. Or even here: https://github.com/alleyinteractive/wordpress-fieldmanager/blob/608cbc18077e21e22533d91db709b6566cbc7903/php/class-fieldmanager-field.php#L288

PR #865, opened 2024-02-29
Issue-863: PHP Deprecated:  htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated

anubisthejackle:
Note: To minimize the impact on the rest of the Fieldmanager fields, this change is localized to the TextArea field, since that appears to be where the issue is occurring.

### Fixes Issue
[PHP Deprecated:  htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated](https://github.com/alleyinteractive/wordpress-fieldmanager/issues/863)

### Description of Change
To address the issue of PHP deprecation errors when a `default_value` field is not set and defaults to `null`, we have modified the logic in the printing of fields within the WP admin area. Specifically, in scenarios where a default value is not set, which leads to a `null` return value, a ternary operation or a direct assignment now ensures that `esc_textarea` receives an empty string instead of `null`. This change prevents `htmlspecialchars()` from throwing a deprecation notice due to receiving a `null` value.

The modification was implemented in the handling of default values within the `Fieldmanager_TextArea` class to circumvent the deprecation notice. It's crucial to emphasize that the default behavior for handling default values remains unaltered to prevent unintended side effects in other areas of the application.

### Use Case Affected
This change impacts any site using fields that extend `Fieldmanager_TextArea`, specifically in scenarios where the `default_value` is not explicitly set. The new default value will be an empty string and not `null` as it was previously.

### Comments/Suggestions
- [@renatonascalves](https://github.com/renatonascalves) has played a crucial role in suggesting the approach of defaulting to an empty string within the `Fieldmanager_Textarea` class to address this deprecation issue.

### Acceptance Criteria
- With no default value set, resulting in a `null` return value, `esc_textarea` will now receive an empty string instead of `null`.
- The approach to handling default values should not be fundamentally altered to avoid unintended consequences in other parts of the application.

### DIFF
--- a/php/class-fieldmanager-textarea.php
+++ b/php/class-fieldmanager-textarea.php
@@ -18,6 +18,13 @@ class Fieldmanager_TextArea extends Fieldmanager_Field {
 	public $field_class = 'text';
 
 	/**
+	 * The default value for the field, if unset.
+	 *
+	 * @var mixed Default value
+	 */
+	public $default_value = '';
+
+	/**
 	 * Construct default attributes; 50x10 textarea.
 	 *
 	 * @param string $label   Field label.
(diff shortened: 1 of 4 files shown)


mogmarsh (approved):
👍

PR #866, opened 2024-02-29
Issue-864: Resolve PHP Deprecations

anubisthejackle:
### Description

While running the unit tests, I run into several PHP deprecations. It'd be great if those issues were resolved since they create noice in logs.

```

PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/wordpress/wp-content/plugins/wordpress-fieldmanager/php/class-fieldmanager-group.php on line 406

PHP Deprecated:  strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/wordpress-develop/src/wp-includes/functions.php on line 7127

PHP Deprecated:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/wordpress/wp-content/plugins/wordpress-fieldmanager/php/class-fieldmanager-field.php on line 992

... more ...
```

Also, consider using `convertDeprecationsToExceptions="true"` in the phpunit.xml so that a test fails if it contains PHP deprecations.

### Use Case

Testing the test suite returns no PHP deprecations.

### Acceptance Criteria

- Add checks to ensure that values are strings before running them through string functions in the three locations identified above
- Update the unit test configuration so that deprecations are converted to exceptions and require resolution before merging a PR


tests/php/test-fieldmanager-checkboxes-field.php
```diff
@@ -7,9 +7,33 @@
...
 class Test_Fieldmanager_Checkboxes_Field extends WP_UnitTestCase {
-	public $post_id;
-	public $post;
-	public $custom_datasource;
+/**
+	 * The post ID of the test post.
+	 *
+	 * @var int
+	 */
+	private int $post_id = 0;
```
kevinfodness, line 15: Note that this changes visibility of the property from public to private, although it should be fine in a test case.

tests/php/test-fieldmanager-context-post.php
```diff
@@ -133,6 +147,8 @@ public function test_context_save() {
 	}
 
 	public function test_programmatic_save_posts() {
+		$this->expectNotToPerformAssertions();
```
kevinfodness, line 150: This test doesn't do anything and should either be updated to perform assertions or removed entirely.

tests/php/test-fieldmanager-richtextarea-field.php
```diff
@@ -565,4 +580,43 @@ public function test_customize_buttons_filter() {
...
+	public function test_default_value_does_not_throw_deprecation() {
+		set_error_handler( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_set_error_handler
+			/**
+			 * Convert deprecations to exceptions. This was removed from PHPUnit in
+			 * PHPUnit 10, so doing so manually here for future compatibility.
+			 *
+			 * @see https://github.com/sebastianbergmann/phpunit/issues/5062
+			 */
+			function ( $errno, $errstr ) {
+				throw new Fieldmanager_Deprecation_Exception( $errstr, $errno );
```
kevinfodness, line 600: Rather than throwing, you could keep track of the number of deprecation messages in a variable in this class and then run an assertion on it later. Tests should have assertions, and that would be a way to bring an assertion into this test rather than relying on it not throwing an exception.

kevinfodness (approved):
A few things to address before merge, otherwise 🍣

PR #898, opened 2026-06-09
fix: exact field-key match in trigger_display_if to prevent false positives

### LINKED ISSUE #873: display_if incorrectly hide siblings when matching their `display-src` values
### Description of the bug

When trying to trigger the show and/or hide fields, `display_if` matches the partial number of a sibling field instead of a full match. 

If you happen to have a display_if logic that share a field name, fields with a partial match ends up getting hidden.

The source of the issue seems to be the use of `.match` here, since it is not a full name match but a partial one (it matches a string based on the field name instead of matching the string wih the field name):

https://github.com/alleyinteractive/wordpress-fieldmanager/blob/584979f3e59b2173fa71633a07c39c0050b11347/js/fieldmanager.js#L299

### Steps To Reproduce

Here is an example you can use to replicate this issue:

```php
function add_custom_meta_boxes() {
	$custom_meta_group = new Fieldmanager_Group(
		array(
			'name'           => 'custom_meta_group', // <--- display_if will match 'meta' in this field name too, even though it should not.
			'serialize_data' => false,
			'children'       => array(
				'meta'   => new Fieldmanager_Checkbox(					[
						'save_empty'    => false,
						'checked_value' => 'meta',
						'label'         => 'Checkbox',
					]
				),
				'text_field' => new Fieldmanager_Textfield(
					array(
						'label'      => 'Text Field',
						'display_if'  => [
							'src'   => 'meta',
							'value' => 'meta',
						],
					)
				),
				'textarea'   => new Fieldmanager_TextArea(
					array(
						'label' => 'Text Area',
						'display_if'  => [
							'src'   => 'meta',
							'value' => 'meta',
						],
						'attributes' => array(
							'rows' => 5,
							'cols' => 50,
						),
					)
				),
				'subject_prefix'        => new Fieldmanager_Select(
					[
						'name'          => 'subject_prefix',
						'label'         => 'Subject prefix',
						'options'       => [
							'none'   => 'None',
							'custom' => 'Custom',
						],
						'default_value' => 'none',
						'display_if'  => [
							'src'   => 'meta',
							'value' => 'meta',
						],
					]
				),
				'subject_prefix_custom' => new Fieldmanager_Textfield(
					[
						'label'       => 'Custom subject prefix',
						'display_if'  => [
							'src'   => 'subject_prefix',
							'value' => 'custom',
						],
					]
				),
			),
		)
	);

	$custom_meta_group->add_meta_box( 'Custom Meta', 'your-post-type', 'normal', 'high' );
}
add_action( 'admin_init', 'add_custom_meta_boxes' );


https://github.com/user-attachments/assets/329717ec-d39c-48ba-b747-9d2c0c394531


```

### Additional Information

I'm working on the assumption that `.match` should match the field name provided in `display-src`, but it is possible there is feature that relies on a loose match.

So this change might require a bit more testing to avoid regressions.

thisismyurl:
## What

`trigger_display_if` used `name.match( displaySrc )` to decide whether a
changed input should trigger visibility changes on `display_if` siblings.
`String.prototype.match()` treats its argument as a regex and matches
substrings, so any trigger input whose full HTML `name` attribute contains
`displaySrc` as a substring incorrectly fires the display logic.

## Reproducer (from #873)

```php
$custom_meta_group = new Fieldmanager_Group([
    'name'           => 'custom_meta_group',  // contains "meta" as a substring
    'serialize_data' => false,
    'children'       => [
        'meta'         => new Fieldmanager_Checkbox(['checked_value' => 'meta']),
        'text_field'   => new Fieldmanager_Textfield(['display_if' => ['src' => 'meta', 'value' => 'meta']]),
        'subject_prefix' => new Fieldmanager_Select([...]),
    ],
]);
```

When `subject_prefix` changes, its input name is
`custom_meta_group[subject_prefix]`. The old code evaluates:

```js
'custom_meta_group[subject_prefix]'.match('meta') != null  // → true!
```

…because `"meta"` is a substring of `"custom_meta_group"`. This causes
`text_field` to be shown or hidden whenever `subject_prefix` changes,
even though `text_field` should only react to `meta`.

## Fix

Replace the partial regex match with an exact bracket-key check:

```js
// Before
if ( name && name.match( $( this ).data( 'display-src' ) ) != null ) {

// After
var displaySrc = $( this ).data( 'display-src' );
if ( name && ( name === displaySrc || name.indexOf( '[' + displaySrc + ']' ) !== -1 ) ) {
```

- `name === displaySrc` handles top-level Fieldmanager fields whose
  input name is the bare field key (no bracket notation).
- `name.indexOf( '[' + displaySrc + ']' ) !== -1` handles nested fields
  where the key is wrapped in brackets (e.g. `group[meta]`). The
  surrounding brackets ensure `[meta]` cannot match `[meta_info]` or
  `[custom_meta_group]`.

## Testing

1. Set up the group from the reproducer above.
2. Change `subject_prefix` — confirm `text_field` visibility does **not** change.
3. Check the `meta` checkbox — confirm `text_field` shows/hides correctly.
4. Run the existing JS test suite: `npm test` (or equivalent).

Fixes #873

---
*Development and testing assisted by AI. All code reviewed and verified manually.*

### DIFF
--- a/js/fieldmanager.js
+++ b/js/fieldmanager.js
@@ -296,7 +296,8 @@ var fm_init = function () {
 		}
 		$( this ).closest( '.fm-wrapper' ).siblings().each( function() {
 			if ( $( this ).hasClass( 'fm-display-if' ) ) {
-				if ( name && name.match( $( this ).data( 'display-src' ) ) != null ) {
+				var displaySrc = $( this ).data( 'display-src' );
+				if ( name && ( name === displaySrc || name.indexOf( '[' + displaySrc + ']' ) !== -1 ) ) {
 					if ( match_value( getCompareValues( this ), val ) ) {
 						$( this ).show();
 					} else {


thisismyurl:
Rebased onto current `main` — branch is fully up to date, no conflicts. CI is showing "action required" rather than running automatically, likely needs a maintainer approval for the workflow run on this fork push.

